PR #10, opened 2014-09-12
Create ssl cert with subject alternative names SAN

Dude4Linux:
The proposed changes add a new utility, tkl-make-ssl-cert.

Usage: tkl-make-ssl-cert [-h|--help] [-o|--out file] [-d|--default] [-w|--wild] [-t|--template file] [-v|--verbose] [-f|--force-overwrite] [-r|--csr] FQDN

# Purpose:

  - Make default certificate for TurnKey GNU/Linux appliance [-d|--default]
  - Make server certificates for Apache/Nginx virtual hosts
  - Make wildcard certificate for entire domain [-w|--wild]
  - Optionally generate a certificate signing request [-r|--csr]
  - Store certificates in /etc/ssl/private because they contain a key
  - Create symlinks in /etc/ssl/certs for backward compatibility

In addition, the fully qualified domain name, FQDN, is determined during firstboot, and the administrator is given the opportunity to override the default value - see ask-fqdn.py.  Selected parameters are shared with other firstboot scripts by leveraging the inithooks.conf file.  This allows the APP_FQDN and APP_IP to be used when generating the ssl subject alternate name (SAN) default certificate and by the ownCloud appliance to configure it's list of trusted domains.

Testing has been done on the core, drupal7, ansible, and owncloud appliances.  Although this is far from an exhaustive test, I feel comfortable that the changes should work with most appliances.  The reason for placing the changes in fab/common, rather than just in products/owncloud is that I feel that other appliances can benefit from having the FQDN set during firstboot.


lirazsiri:
Did you get around to fixing the issue with the certificate? Can I merge?

Dude4Linux:
Liraz, I would hold off on this merge.  In fact, I should probably remove the request.  The issue was that I had problems getting OSX Safari and Chrome/Chromium to accept the certificates.  I think I have that figured out but I decided to backoff and take a less aggressive approach.  I started work, but then got sidetracked on other projects and haven't had time to finish.  I've also been trying to absorb the material from Ivan Ristic's book "Bulletproof SSL and TLS".  Can you give me a time-frame as to when you want to make the next release?

PR #11, opened 2015-02-17
make vim.tiny alternative for vim (instead of symlink)

JedMeister:
Closes [#300 ](https://github.com/turnkeylinux/tracker/issues/300) 

IMO a neater way of making vim.tiny open with `vim` command in TurnKey but will automatically update `vim` to point to full vim (aka vim.basic) if installed.

I have tested this both in a pre-installed Core; but also by using my patched `common` to build a new Core iso (just to be 100% sure) and it works sweet! :) `vim` launches vim.tiny as per current behaviour, but if full vim is installed (`apt-get install vim`) `vim` command will launch vim.basic (as users would expect).

What do you think?

### DIFF
--- a/conf/turnkey.d/vim.tiny
+++ b/conf/turnkey.d/vim.tiny
@@ -1,6 +1,6 @@
 #!/bin/sh
 
 VIMTINY=$(which vim.tiny)
-[ "$VIMTINY" ] && ln -s $VIMTINY /usr/local/bin/vim
+[ "$VIMTINY" ] && update-alternatives --install /usr/bin/vim vim /usr/bin/vim.tiny 10
 
 exit 0


Dude4Linux:
Is there any chance of replacing vim.tiny with vim.nox?  (apt-get install vim-nox)  That is usually the first thing I do after installing a new host.  I also change my vimrc defaults, but that's another matter.  I just can't live with vim.tiny mangling cursor control chars when accessed through ssh.  FYI try editing a line and then moving to a different line without hitting esc.  vim-nox has all the fancy features of vim.basic without the X-window support.

lirazsiri:
John, replacing vim-tiny with vim-nox would add significantly to the size of Core. 18MB of archives, 32MB decompressed. Since it's so easy to install vim-nox via the package management I'm not sure that would be the best default.

Dude4Linux:
That's certainly your call.  I already have a script that I use to customize each appliance I setup and vim-nox is only one of the things I like to tweak. :)

PR #12, opened 2015-04-17
Fix dpkg-vendor output

JedMeister:
closes issue #196 (https://github.com/turnkeylinux/tracker/issues/196)

I haven't tested bt building core but I have tested by applying to a running system


alonswartz:
Is this change still required if the distribution doesn't manage the entire package repository? For example, Ubuntu sync the entire package archive and their sources.list only points to the Ubuntu package archive. IIRC, they sync the source packages and build the packages themselves. On the flip side, TurnKey's sources.list points directly to the Debian package archive as well as the TurnKey package archive which contains TurnKey specific packages.

Additionally, does this change effect any other output except `dpkg-vendor --query vendor`?

JedMeister:
It's not 'required' as such but it is 'requested'/'recommended' (by Debian) for all Debian derivatives. TBH I had forgotten about this...

TL; DR - last paragraph...
TBH there has been some discussion lately WRT this change and someone on the mailing list (a guy from Tails perhaps?) said that it broke some package installs so they changed it back. Response from the Debian guy (who appears to be essentially in charge of maintaining the derivatives stuff) was that any packages that fail because of this should have bugs lodged against them...

Long and short is that I propose that we put this one on ice for now. I think that there are plenty to be done re update to Jessie to risk adding extra headaches... Let's just close this?!

alonswartz:
OK, closing for now...

JedMeister:
I'm reopening this for discussion.

JedMeister:
As per discussion https://github.com/turnkeylinux/tracker/issues/196 I'm going to merge this now (for v15.0). 

Hopefully we won't hit any issues. If we do, we'll try to address them as we go. 

Rolling this back is an option in the unlikely event that things go really pear shaped. FWIW, I've added this change to my Stretch laptop and have yet to see any negative impact.

PR #15, opened 2015-05-12
fixes for v14 rails appliance

wvengen:
This should fix rails build problems on the v14.0rc1 tkldev appliance.

* builds succesfully
* boots
* live cd: mysql is not started by default (needs `service mysql restart`)
* live cd: passenger segmentation fault at 0x0 in libv8 / therubyracer


wvengen:
I can't seem to fix the passenger runtime segfault in libv8. The following doesn't seem to help:
* using Debian's `libapache2-mod-passenger` instead of the passenger gem
* using [system v8](http://stackoverflow.com/questions/23536893/installing-libv8-and-therubyracer-gems-on-mavericks-with-system-v8-installation)

JedMeister:
Regardless it's a good start and I imagine will provide at least hints for further work...

Have a look @qrntz :smile:

alonswartz:
Closed per inclusion in #18

wvengen:
Thanks!

PR #18, opened 2015-05-14
Rails update to v14

qrntz:
This incorporates fixes from https://github.com/turnkeylinux/common/pull/15 and additionally gets rid of the mentioned segfault by replacing the JS interpreter for ExecJS (therubyracer → Node.js)

Actually, this is not a workaround but fairly good in itself (see https://github.com/sheerun/rails4-bootstrap/issues/18).


wvengen:
I can confirm that this works. After running `service mysql restart` manually from the live cd, I can access the website properly. (Although I did see `therubyracer` and `libv8` mentioned in the `bundle install` step _after_ the one run as part of `rails new`.)

qrntz:
MySQL works for me (is started automatically on first boot).

About the gems, good call! I forgot to double-check the actual base appliance, which echoed therubyracer to the Gemfile (fixed in 23a3dfb53b23ba3c428bdfe2adf2d5c36424c89e).

PR #30, opened 2015-05-31
Add oracle install script

JedMeister:
What do you think @alonswartz? 

Personally I quite like the idea, although I generally have well founded reservations about using a repo designed for Ubuntu in this case I don't think that it's too big a deal...

I was going to adjust the code slightly but figured that I probably should check on your opinion before I spend any time on it...

Also wasn't sure whether it should be in all Tomcat appliances or just Tomcat & Tomcat-Apache?


alonswartz:
@JedMeister - What is the use-case? I'm leaning towards having this as documentation...

JedMeister:
@SiKing made it as he needed to use Oracle Java on a TKL Tomcat based server for a client. He shared it as he thought that it might be of value. 

You're probably right. Having it as documentation is a good way to go.

It's now on the wiki: http://www.turnkeylinux.org/doc/how-to-install-oracle-java

PR #33, opened 2015-06-09
Add ncurses-term to base plan

qrntz:


alonswartz:
@qrntz can you provide rational and/or use case?

qrntz:
@alonswartz, sure, support for widespread $TERM values not included by default (e. g. rxvt-*, eterm, konsole and flavors of xterm among others as per http://packages.ubuntu.com/trusty/all/ncurses-term/filelist). I find myself having to install ncurses-term as part of every appliance because of the default terminfo directory not having rxvt-unicode (which means no ^L to clear buffer and no ncurses output silently) and I'm sure it would be useful to others likewise. Fairly small package, too (2239 kB).

PR #34, opened 2015-06-18
Changed order of SSLCert Generation

OnGle:
Changed the order of ssl cert generation as to
stop programs that set hostname after sslcert regened
from encountering issues.


alonswartz:
This change requires updating the BLACKLIST in [1]. Though, the appliances which require the hostname (eg. magento) might fail when re-running [1] if it is skipped, and the hostname is changed.

If the hostname is the issue, maybe we should move the hostname inithook before the ssl-certs?

Another alternative might be to update the affected appliances to handle the error, and do the right thing (ie. re-run the ssl-cert inithook)?

I'm open to ideas. We just need to make sure the solution is idempotent, and will do the right thing in different scenarios (if a service is using the cert, we might need to restart it. if the user setup a signed-cert, we don't want to replace it with a self-signed cert, ...)

[1] https://github.com/turnkeylinux/inithooks/blob/master/turnkey-init

alonswartz:
@JedMeister any update on this per our phone discussion?

JedMeister:
After testing and testing and testing some more I have been unable to reproduce this issue. As I said over on turnkeylinux/tracker#400 I'm not sure what's going on... But it works now so we can close this...

PR #35, opened 2015-06-24
Add support in common for Adminer via lighttpd

DocCyblade:
This also combines the apache config as well. See code for details and issue #387 for running comments 

I also have one with just adminer-lighttpd 
(https://github.com/DocCyblade/common/tree/adminer-lighttpd)

If you would rather not combine them just reject this one and I will submit pull request on that one instead.

Once merged, I will finish postgres and mysql apps to use this.


JedMeister:
Great work Ken. Unless @alonswartz can see something that I have missed I think that this is good to go :+1: 

Like I said on the issue thread, I like having this all-in-one (i.e. Apache and LigTTHPd conf together). It is the way that Debian packages are going now and with the small size of text files it seems like a nice way to simplify things in common (worst case appliance can remove the one they don't need if desired). 

Also by adding the Apache.conf but not yet removing the separate one it means that nothing will break and we can then transition to removing that later...

alonswartz:
Nice! 

In general, I think it makes sense to separate the webserver specific overlays, just as you separate the webserver specific confs. But, I haven't been involved in this discussion so I am most likely missing the rational, in which case I am deferring to you guys best judgement and merging.

JedMeister:
As I hinted above, Debian have set a precedent of now including multiple webserver conf scripts (where available) so that they all sit in /etc/appname/; e.g. /etc/appname/apache.conf /etc/appname/lighttpd.conf /etc/appname/nginix.conf then the user symlinks the relevant one. For the sake of a redundant tiny text file or two it seemed to me like a reasonable way of doing things.

As you guessed, Ken and I discussed it and agreed that a small redundant text file was a small price to pay for one less entry in common (i.e. reduces the noise/complexity in common). 

If you disagree then we can easily remove it and make a separate entry...

PR #39, opened 2015-07-10
Adminer updates for db selection for lamp and lapp and others

DocCyblade:
Adds conf scripts to configure adminer to specific database mysql or pgsql and sets default database to localhost. Will also use these for mysql and postgres appliances as well. 

Also changes lamp/lapp make files to use new adminer changes

### DIFF
--- a/conf/adminer-mysql
+++ b/conf/adminer-mysql
@@ -1,0 +1,25 @@
+#!/bin/sh -e
+
+# adminer-mysql
+#
+#   common conf script to configure adminer to only show mysql
+#   database option and localhost for server.
+#   Also removes version checking code
+
+ADMINER_DIR=/usr/share/adminer/adminer/include
+ADMINER_BOOTSTRAP=$ADMINER_DIR/bootstrap.inc.php
+
+
+# Comment out uneeded db options
+sed -i '/^include "..\/adminer\/drivers\/sqlite.inc.php"/ s|include|\/\/ include|' $ADMINER_BOOTSTRAP
+sed -i '/^include "..\/adminer\/drivers\/pgsql.inc.php"/ s|include|\/\/ include|' $ADMINER_BOOTSTRAP
+sed -i '/^include "..\/adminer\/drivers\/oracle.inc.php"/ s|include|\/\/ include|' $ADMINER_BOOTSTRAP
+sed -i '/^include "..\/adminer\/drivers\/mssql.inc.php"/ s|include|\/\/ include|' $ADMINER_BOOTSTRAP
+
+
+# Set localhost as default database
+sed -i '/^define("SERVER", \$_GET\[DRIVER\])/ s|\$_GET\[DRIVER\]|"localhost"|' $ADMINER_BOOTSTRAP
+
+
+# Remove version checking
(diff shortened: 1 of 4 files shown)


DocCyblade:
This should close https://github.com/turnkeylinux/tracker/issues/360 

Also will use this in issue https://github.com/turnkeylinux/tracker/issues/386 and https://github.com/turnkeylinux/tracker/issues/385

JedMeister:
Great work Ken! :+1: 

My only thought is that perhaps these would be better called from the relevant DB mk files (i.e. [mysql.mk](https://github.com/turnkeylinux/common/blob/master/mk/turnkey/mysql.mk#L2) and [pgsql.mk](https://github.com/turnkeylinux/common/blob/master/mk/turnkey/pgsql.mk#L2))?

Thinking aloud... Then they would auto apply to all LAMP (via [lamp.mk](https://github.com/turnkeylinux-apps/nginx-php-fastcgi/blob/master/Makefile#L7)), LAPP (via [lapp.mk](https://github.com/turnkeylinux/common/blob/master/mk/turnkey/lapp.mk#L7)) and other webserver appliances (e.g. [lighttpd](https://github.com/turnkeylinux-apps/lighttpd-php-fastcgi/blob/master/Makefile#L7) and [nginx](https://github.com/turnkeylinux-apps/nginx-php-fastcgi/blob/master/Makefile#L7)) as well as the standalone [MySQL](https://github.com/turnkeylinux-apps/mysql/blob/master/Makefile#L7) & [PostgreSQL](https://github.com/turnkeylinux-apps/postgresql/blob/master/Makefile#L6) appliance - all without being specifically called...

If we were to go that way though, we'd need to ensure that it checks for Adminer (and the specific files existence) first. Otherwise in some cases (if app has DB but not Adminer) it will cause the build to fail. E.g. Django has a MySQL DB backend but does not have Adminer (see [here](https://github.com/turnkeylinux-apps/django/blob/master/Makefile)).

@alonswartz - What are your thoughts on this? Better to follow my suggestion or add to lamp.mk, lapp.mk in common and mysql, pgsql, lighttpd and nginx apps individually?

DocCyblade:
I had thought of that, should have asked. It would make more sense as any app using mysql or pgsql would get it. If we do this, would we need to test every app that uses these to ensure no issues.

JedMeister:
IMO it's the way to go, but I'd really like to hear from Alon first as he may have some really valid reason why not that I haven't thought of...

We need to test all the apps anyway! :smile: in our QA round (which has started but only LAMP appliances tested so far so it won't have negative effect on them...).

DocCyblade:
Let me know if you think it be best to go with the sql and pgsql make files instead. I have a local branch I have been working on that I can use instead of this one, or I can just merge this branch. Just let me know how you want to proceed

JedMeister:
Sorry for delay on this one Ken...

TL;DR - we'll merge this one as is!

Alon and I spoke at length on this one today. The final decision (for now at least) is to err on the side of caution and make the Adminer DB specific change explicit. So this PR should be merged and Adminer will be auto configured for any appliance that includes lamp.mk or lapp.mk. Any other appliances that need it will need to explicity call the relevant adminer conf in their Makefile...

JedMeister:
Great work Ken! :+1: I just tested these with Lighttpd and Nginx servers. Works a treat so I've added it to their Makefiles too! :smile:

PR #43, opened 2015-08-04
Add SambaDAV & NFS support to File Server appliance

qrntz:
Fixes https://github.com/turnkeylinux/tracker/issues/20


alonswartz:
@qrntz do any other appliances use nfs and/or sambadav besides the fileserver? shouldn't these changes be made in the fileserver appliances?

qrntz:
@alonswartz, good call, but I tried to mirror the *status quo*, which had confs like ```fileserver-storage``` and overlays like ```samba-fileserver``` in common (which nothing else uses to my knowledge, not even the domain controller one). I also imagine the fileserver appliance will serve as a base for a future Pydio one, so ```fileserver.mk``` is useful to have in common, referring to the appropriate confs/overlays also placed in common.

alonswartz:
Valid points, merging...

PR #44, opened 2015-08-11
Add tmux to core (closes #31)

JedMeister:
closes turnkeylinux/tracker/issues/31

### DIFF
--- a/plans/turnkey/base
+++ b/plans/turnkey/base
@@ -72,6 +72,8 @@ bsd-mailx
 postfix
 webmin-postfix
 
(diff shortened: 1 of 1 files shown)


alonswartz:
Personally I love tmux, not sure how I ever lived without it. But, when if/when we include it by default I think we should do so with a tweaked configuration.

Additionally, we might want to look into auto-starting/attaching to a session, though I recall there were multiple issues when we did this with byobu in one our RC's, which ultimately led to dropping the idea completely.

PR #42, opened 2015-08-04
Security hardening

Dude4Linux:
References:
[Guide to Deploying Diffie-Hellman for TLS](https://weakdh.org/sysadmin.html)
[Configuring Apache, Nginx, and OpenSSL for Forward Secrecy](https://community.qualys.com/blogs/securitylabs/2013/08/05/configuring-apache-nginx-and-openssl-for-forward-secrecy)
[Strong SSL Security on Apache2](https://raymii.org/s/tutorials/Strong_SSL_Security_On_Apache2.html)
[Logjam: the latest TLS vulnerability explained](https://blog.cloudflare.com/logjam-the-latest-tls-vulnerability-explained/)
[BREACH mitigation](https://community.qualys.com/message/20404#20404)
[Configuring Apache for Perfect Forward Secrecy](http://www.lorrin.org/blog/2013/07/03/configuring-apache-for-perfect-forward-secrecy/)
[Eveho Blog - SSL/TLS, ciphers, perfect forward secrecy and Tomcat](https://blog.eveoh.nl/2014/02/tls-ssl-ciphers-pfs-tomcat/)
Testing:
[testssl.sh](https://github.com/drwetter/testssl.sh) - use git clone to install on your scanning host
nmap - $ sudo nmap -p 10000-20000 -sSV -T4 <hostname>
[OpenVAS](http://www.openvas.org/) - available on [Kali Linux](https://www.kali.org/)

PR #48, opened 2015-08-26
SSL hardening

JedMeister:
This is a continuation and reconfiguration of John's fantastic work on SSL/TLS security hardening. 

Whilst the theory of centralising everything is nice; in practice seeing as how webservers etc configuration mechanisms and options are quite different I have moved significant amounts of what was scripted into overlays as IMO it makes for easier reading and future tweaking. It also removes any need for a firstboot script to reconfigure site files... However the ciphers are still implemented with my modified version of John's script as IMO it makes tons of sense to use a consistent cipher list across all webservers/etc and have it set in just one place.

This PR includes the full git history of John's work plus some input from others...

Please merge this and close PR #42 

This also closes issue turnkeylinux/tracker/issues/439


alonswartz:
Excellent work @Dude4Linux @JedMeister @qrntz !!

PR #54, opened 2015-10-30
Add dtach to base plan, call it instead of screen for turnkey-init-fence

qrntz:


alonswartz:
IIRC, @lirazsiri already implemented this, so I'm closing...

qrntz:
Yep.

PR #56, opened 2015-12-20
Lock Rack version to 1.6.4 to match Jessie provided Ruby

qrntz:


JedMeister:
Makes sense...

alonswartz:
Aren't the later versions backwards compatible with the ruby jessie version?

If we really need to pin to a specific version, it might be a better idea to have a debian_version mapping to rack_version. This way, when we move to the stretch, the build will either break or will install the latest version (instead of the hardcoded jessie version).

qrntz:
@alonswartz, apparently they're not. The dependency resolver for ```gem``` tries to pull in Rack of a version newer than the Ruby version in Jessie supports which results in a fail.

You raise a valid point. Not sure how best to detect the Debian version from conf though.

alonswartz:
`fab/share/product.mk` exposes a few useful variables, one being `CODENAME` (for example: jessie). You can see an example of this in https://github.com/turnkeylinux/common/blob/master/conf/turnkey.d/apt

qrntz:
Thanks Alon. Does this look OK?

alonswartz:
@qrntz conf/rails looks good, but conf/rails-pgsql needs to be updated as well doesn't it?

qrntz:
Sure.

PR #57, opened 2015-12-25
turnkey-make-ssl-cert: use hostname only if there are no FQDNs available

qrntz:
Because what else can we use for the CN? The IP is a bigger information disclosure than the hostname since it could leak network topology.
Without this, openssl called from turnkey-make-ssl-cert fails on empty CN.

### DIFF
--- a/overlays/turnkey.d/sslcert/usr/local/bin/turnkey-make-ssl-cert
+++ b/overlays/turnkey.d/sslcert/usr/local/bin/turnkey-make-ssl-cert
@@ -172,6 +172,7 @@ create_temporary_cnf() {
     done
   else
     [[ "$args" == "" ]] && args="$(hostname -A)"
+    [[ "$args" == "" ]] && args="$(hostname)"
     for fqdn in $args
     do
       add_san "$fqdn";                      # fqdn


Dude4Linux:
@qrntz Good catch!!  CN should never be empty.  I made the bad assumption that hostname -A would return the short name if a fqdn wasn't available.  I never caught the problem because in my test setup with Proxmox, my test vm's are permanently setup in dhcp and dns so a fqdn is always available.  I think this is another instance of the race condition where there is a short delay between dhcp request and the availability of the fqdn via host name resolution.  Perhaps we need to add the wait-for-host routine here as well.

I've been including the line

    export APP_DOMAIN=$(hostname -A)

in my inithooks.conf.  I'm guessing the same problem exists here.  I will try adding the wait-for-host and test for empty fqdn into inithooks.conf.

qrntz:
@Dude4Linux, as far as I can see, a hostname was picked unconditionally previously but 7a2ed82889f7a9fdc19f1b53f18eec6d9b52d242 changed that.

I think doing it this way is a bit better since it doesn't leak the hostname if a FQDN *is* available.

My observation wasn't based on a race condition, but of course watching out for one would be good.

Dude4Linux:
@qrntz Aahh, I missed that change.  Must have happened about the time I left for Europe.  Most of the options I added to turnkey-make-ssl-cert came from work I was doing trying to add Plex to ownCloud to create a media-server appliance.  I never got the combination to work well together.  Recent versions of ownCloud have added a trusted_domains array.  This array needs to be populated with any host or domain name that may be used to connect to the server.  This includes localhost and in some cases the IP address(s).  In order to provide a seamless experience for the users with ownCloud clients on Windows, Linux, OSX, iOS and android, I found it was necessary to include the same list of domains as SANs in the certificate.  I had intended the --ip option only for this purpose.  Since I was thinking in terms of a home network, I didn't give much thought to the possibility of leaking information.

Dude4Linux:
Note for future reference: In my testing of the LXC appliance, I found that 'hostname -A' can't be relied on to return the correct FQDN.  In certain tests I found it was returning the FQDN of one of the lxc containers instead of the host itself.  I ended up using 

    FQDN=$( host $(hostname) | head -1 |  cut -d ' ' -f1 )

The 'head -1' is needed to handle the case where the FQDN is defined by a cname entry in dns and multiple names are returned.  In this case, we only want the first name returned.

JedMeister:
Hi guys. 

@qrntz - How about rather than adding a new commit we just revert https://github.com/turnkeylinux/common/commit/7a2ed82889f7a9fdc19f1b53f18eec6d9b52d242? That would resolve this issue wouldn't it?

qrntz:
@JedMeister, sure would but I believe @lirazsiri wanted to stop it from leaking the product name and if we can't do that entirely, might as well settle for the next-best solution.

My version won't record the hostname if a FQDN is available, but will do it otherwise.

Dude4Linux:
@JedMeister, I originally created the turnkey-make-ssl-cert while I was working on a Plex mediaserver based on the ownCloud appliance. I found there were some issues with ownCloud and getting it's certificates accepted by all the clients, i.e. Windows, Linux, Android, OSX, and iOS.  This is a different use case than for other appliances.
For the v14.1 version, I'm in favor of following @qrntz Anton's suggestion of including only the fqdn in the default cert with fallback to the hostname if fqdn can't be determined.  I would also like to add an option to include all the possible ways of addressing the appliance in the SAN list for those special cases where it is helpful.

JedMeister:
Ok. You guiys have convinced me! :smile:

PR #58, opened 2015-12-26
pgsqlconf.py: su into a login shell

qrntz:
Prevents the error about being unable to change directory to /root on su invocation.


JedMeister:
:+1:

PR #59, opened 2015-12-29
Install composer on LAMP/LAPP by default

qrntz:
Resolves https://github.com/turnkeylinux/tracker/issues/527


JedMeister:
:+1:

alonswartz:
@JedMeister what is the status on this? Should we hold off on merging since you reassigned to 14.2?

JedMeister:
Yes good call Alon. It'd be good to tidy up all the appliances (that install composer in the app repo) at the same time. That way it's more likely that things won't be forgotten or missed.

JedMeister:
:+1:

PR #61, opened 2016-01-23
Make Samba handle sudoadmin

qrntz:
* sets the Samba admin user to «admin»
* sets the password
* deactivates the root user

Not sure if the number is right. Also, this is basically a version of https://github.com/turnkeylinux/common/blob/master/conf/samba-rootpass which proved to be sufficient.

### DIFF
--- a/overlays/samba-fileserver/usr/lib/inithooks/firstboot.d/35samba-sudoadmin
+++ b/overlays/samba-fileserver/usr/lib/inithooks/firstboot.d/35samba-sudoadmin
@@ -1,0 +1,16 @@
+#!/bin/bash -e
+# adapt samba to admin user
+
+[ -n "$_TURNKEY_INIT" ] && exit 0
+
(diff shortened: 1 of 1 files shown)


JedMeister:
Looks good to me :+1:

JedMeister:
@alonswartz - Can you please have a quick look and merge this. 

Also it closes https://github.com/turnkeylinux/tracker/issues/523

alonswartz:
@qrntz wouldn't it make sense to have just one samba-pass script which does the right thing instead of having two (rootpass and sudoadmin)?

We also might want to remove the unlocking of root in samba-rootpass?

alonswartz:
I'm merging the PR, then we'll handle the proposed changes/fixes...

alonswartz:
Sorry about that, please ignore my previous comments. I was mistakenly under the impression that the samba-rootpass was an inithook. We should be good to go....

PR #62, opened 2016-02-22
added webmin-filemin to base plan

JedMeister:


alonswartz:
@JedMeister shouldn't webmin-filemin be replacing webmin-file? Do we need both?

JedMeister:
I vote we keep both for now and remove the old one next Webmin release. It is deprecated but Webmin have renamed the old file manager "java file manager" in the UI (so there is no conflict).

PR #64, opened 2016-02-25
Fixed turnkey pylib git.py and GitPython clash

OnGle:
Added virtualenv to /var/www/py2web/venv
Installed GitPython to virtualenv

Added 2 lines to enter virtualenv after system python starts
py2web into wsgihandler.py


OnGle:
closes https://github.com/turnkeylinux/tracker/issues/578

PR #76, opened 2016-09-05
Changed inithooks to auto generate mysql password

OnGle:
Changed inithooks to auto generate password as well as added confconsole plugin to generate/view generated password as well as set password manually (cannot view manually set passwords)

Relies on confconsole "plugins"


JedMeister:
Thanks for doing this work @OnGle - but as per discussions we had ages ago, following further discussions, we decided to put this on hold. I've now just decided to close it completely.

PR #81, opened 2017-02-10
Adminer 4.2 update to common

DocCyblade:
As the title says. This will close https://github.com/turnkeylinux/tracker/issues/734


JedMeister:
Awesome work! So you didn't need to tweak any of the appliances at all then!? That's very cool! 

PS I'll have a closer look at this next week, and merge then. Cheers.

PR #82, opened 2017-02-27
Set lvm/no_boot debconf option in installed rootfs

qrntz:
Fixes one half of https://github.com/turnkeylinux/tracker/issues/782


qrntz:
This is not needed anymore.

PR #91, opened 2017-06-07
ruby 2.3.3 with ruby-build in rails

qq7:
Seeing this [comment](https://www.turnkeylinux.org/comment/27037#comment-27037)  I've decided to look up a way to install newer Ruby version.

It does not provide a way to switch among Ruby versions or create development environments. However, it installs ruby 2.3.3, which allows to use newest gems versions (like rails >= 5)

### DIFF
--- a/plans/turnkey/rails
+++ b/plans/turnkey/rails
@@ -5,9 +5,6 @@ apache2-mpm-prefork
 mysql-server
 webmin-mysql
 
-ruby
-ruby-dev
-
 build-essential            /* needed to build gems */
 libmysqlclient-dev         /* needed to build mysql gem */
 libcurl4-openssl-dev       /* needed to build passenger apache2 module */


JedMeister:
Nice one. Thanks @qq7 ! :+1: I'll have a look over it tomorrow (it's 10pm here now).

conf/rails
```diff
@@ -16,20 +16,14 @@ LOGSDIR=/var/log/$APPNAME
 [ "$FAB_HTTP_PROXY" ] && export HTTP_PROXY=$FAB_HTTP_PROXY
 GEMOPTS="--no-rdoc --no-ri"
 
-gem update --system
+git clone https://github.com/rbenv/ruby-build.git /usr/local/src/ruby-build
```
JedMeister: As a general rule, unless there's a good reason not too, shallow clones are best. I.e. `--depth 1`

JedMeister:
Otherwise, it looks pretty good to me. Although as has been mentioned, Ruby isn't a strength of mine, so I'll ask @qrntz to have a double check as he would be more aware of what may cause issues with other appliances.

qrntz:
I'll do that.

qq7:
https://github.com/turnkeylinux/common/pull/95

PR #65, opened 2016-03-06
Fileserver migration to WebDAV CGI

qrntz:

PR #93, opened 2017-06-10
migration to WebDAV CGI - Common

DocCyblade:
Since PR https://github.com/turnkeylinux/common/pull/65 was almost a year old and a lot has changed, I have taken his changes and got it to work with v1.1.2. I have built this with file server and it works like a treat. I did try media server but it did not work. I am looking into that. If you want to wait I am ok with that. I think it's because media server was still using the old web client.


DocCyblade:
Looks like I also need to look at PR https://github.com/turnkeylinux-apps/fileserver/pull/3 from file server and PR https://github.com/turnkeylinux-apps/torrentserver/pull/8 for torrent server!

DocCyblade:
It seems I did pull in Anton's changes for file server [on my working file server branch](https://github.com/DocCyblade/fileserver/commits/webdavcgi) and it indeed works. I can confirm that PR https://github.com/turnkeylinux-apps/fileserver/pull/3 still works with the updated v1.1.2 of WebDAVCGI

PR #95, opened 2017-06-18
ruby v2.3.3 with rbenv for Rails-based appliances

qrntz:
Credit to @qq7 for the initial implementation.


JedMeister:
Looks good. Thanks @qrntz and @qq7.

JedMeister:
PS, Perhaps we should just leave GitLab and Canvas for now? And look at updating the `rails-pgsql` common script later? I've actually already built GitLab to v14.2 anyway...

PR #114, opened 2018-04-02
Hardening for version 15.0

Dude4Linux:
These changes implement many but not all of the system hardening recommendations made by Lynis and OpenVAS.  Some recommendations such as disabling root login and using an admin account with sudo will probably never be implemented. Others such as changing the default umask from 022 to 027 might lead to problems in certain appliances where scripts assume the default values.  I tried to make the sysctl variables easy to override either in /etc/sysctl.conf or by any file beginning with a number greater than 10 in /etc/sysctl.d/.
One thing to remember is that in a container (LXC or LXD), certain sysctl variables are read-only in the container and cannot be overridden by these settings. In this case, the variables must be overridden in the container host.  Depending on the host, hardware, VM, AWS, etc. it may not be possible to achieve a perfect Lynis score, and this is perfectly acceptable.

### DIFF
--- a/conf/turnkey.d/postfix-local
+++ b/conf/turnkey.d/postfix-local
@@ -8,6 +8,9 @@ fi
 postconf -e inet_interfaces=localhost
 postconf -e myhostname=$HOSTNAME
 
+# harden postfix banner
+postconf -e smtpd_banner='$myhostname ESMTP'
+
 /etc/init.d/postfix start
 /etc/init.d/postfix stop
 
--- a/conf/turnkey.d/roothome
+++ b/conf/turnkey.d/roothome
@@ -1,5 +1,10 @@
 #!/bin/bash -e
 
+# harden ssh client directories
+mkdir -m 0700 /etc/skel/.ssh
+umask 0077 && touch /etc/skel/.ssh/.anchor
+cp -dRn /etc/skel/.ssh /root
+
 cp /etc/skel/.bashrc /root
 [ -d /etc/skel/.bashrc.d ] && cp -dRn /etc/skel/.bashrc.d /root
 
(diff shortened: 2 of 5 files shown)


Dude4Linux:
@JedMeister While I'm thinking of it, I added some files related to hardening ssh to the /etc/skel . You will need to check to see if they cause an issue with AWS. The reason for putting them in skel, was to assure that they get applied to all user accounts as well as root.

JedMeister:
@Dude4Linux - thanks for this. Apologies on my delayed response. I've been dealing with a few other issues that I need to get knocked out of the way first... 

Anyway, I've just had a look through and on face value this looks great. 

When I initially say the SSH stuff, I was not so sure about `MaxAuthTries 2`, but after having a quick read of the docs, I think that's a good idea. As it just drops the connection (doesn't actually block you), that will integrate well with the v15.0 inclusion fail2ban (it's initially just an MVP - only covering SSH; but it's a start; with an aim to cover all "admin" type interfaces in v15.1).

I'm going to give this a bit of a test drive and see how things go. Out of interest, have you done much testing of them yourself? I'm assuming you have, but curious... Regardless, thanks as per always John! :smile:

Dude4Linux:
@JedMeister - Now it's my turn to apologize. I've been on the road for the past couple of weeks.  I got bit by the following bug, https://bugs.launchpad.net/ubuntu/+source/intel-microcode/+bug/1759920.  Just managed to get my laptop functioning again which makes it easier to communicate.  For awhile, I was afraid the problems were the result of something I had done.  I'm happy to report that the problems were the result of the aforementioned bug and not anything to do with the system hardening or the PDE.

PR #116, opened 2018-04-07
add root-like user

spaghettimaster:
closes https://github.com/turnkeylinux/tracker/issues/954

### DIFF
--- a/conf/mysql
+++ b/conf/mysql
@@ -19,7 +19,7 @@ DROP DATABASE IF EXISTS test;
 DELETE FROM db WHERE Db='test' OR Db='test\\_%';
 
 # add root-like user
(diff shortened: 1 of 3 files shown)


JedMeister:
Sorry to do this to you @spaghettimaster but could you please make a minor change to this and make the username `adminer` (rather than `admin`)?

I think that will reduce the possible confusion for AWS Marketplace users. As background, because of AWS Marketplace policy; for TurnKey appliances on AWS Marketplace, the `root` Linux user account is disabled and a new Linux user `admin` is created and used instead.

On reflection, I'm concerned that using the username `admin` for MySQL may create additional confusion for those users. If we use `adminer` then it's really explicit that it's different.

spaghettimaster:
Hey @JedMeister,  its done!
I made another commit.

JedMeister:
Thanks mate, you rock! :smile: 

PS, sorry I haven't merged your last few PRs yet. There's nothing wrong with them from what I can see, it's just that I'm getting close to merging everything into master, so have left these last few PRs to deal with once I've got everything else updated.

PR #126, opened 2018-05-19
Change TCPKeepAlive to yes

qrntz:
This mimics TKL v14.x behavior. Re security, consider:

https://unix.stackexchange.com/questions/208055/is-sshd-config-tpckeepalive-still-using-unencrypted-channel-and-therefore-vulner

Closes https://github.com/turnkeylinux/tracker/issues/1092

### DIFF
--- a/conf/turnkey.d/sshd
+++ b/conf/turnkey.d/sshd
@@ -19,10 +19,12 @@ sed -i '{
 
 cat >> /etc/ssh/sshd_config <<EOF
 
+# https://github.com/turnkeylinux/tracker/issues/1092
+TCPKeepAlive yes
+
 # SSH hardening recommended by lynis
 UsePrivilegeSeparation sandbox
 X11Forwarding no
-TCPKeepAlive no
 AllowTcpForwarding no
 ClientAliveCountMax 2
 MaxAuthTries 2


JedMeister:
@Dude4Linux - please note this reversion of a change you implemented for v15.0. See the link (above) provided by @qrntz for security implications. (IMO allowing the connection to time out is preferable).

Dude4Linux:
@JedMeister -- Anton @qrntz didn't give a reason for wanting to restore the TCPKeepAlive back to the 14.x default `yes`.  It's probably because he has a use case where ssh connections timing out are inconvenient.
I looked for reasons why CISOfy's `lynis` is recommending changing the setting to `no` and found this discussion thread, see https://github.com/CISOfy/lynis/issues/120.  Apparently they made the change back in 2016.  I checked US-CERT and verified that there is no current warning re: TCPKeepAlive.

Eric Light @rhyven says

> Thanks for bringing that to my attention, @rilindo. You're quite right. It seems that there's no actual exploitable vulnerability with TCPKeepAlive(1) -- yet. But let's fix this.
> 
> After a bit of reading this morning (thanks again), I've found:
> 
> > TCPKeepAlive: This option only uses TCP keepalives (as opposed to using ssh level keepalives)
> 
> vs
> 
> > ServerAliveInterval: The server alive messages are sent through the encrypted channel and therefore will not be spoofable. The TCP keepalive option enabled by TCPKeepAlive is spoofable.
> 
> (Ref: man page for ssh_config)
> 
> It appears much more appropriate for Lynis to:
> 
>  - Ensure  TCPKeepAlive is set to No
>  - Suggest that ServerAliveInterval could be set to a value greater than zero (2)

More information about `ServerAliveInterval` and `ServerAliveCountMax` can be found here https://drupal.star.bnl.gov/STAR/comp/sofi/facility-access/ssh-stable-con

Although the two mechanisms behave very differently, I think that setting `ServerAliveInterval` and `ServerAliveCountMax` may have the desired effect of keeping an idle `ssh` connection open as long as the `Server` is still alive and dropping connections when the `Server` fails to respond to `ServerAliveCountMax` queries.  The suggested values of 15 minutes for Interval and 3 for CountMax might be a starting point for testing.

JedMeister:
Hey @Dude4Linux

Actually, the reason is the opposite. In my experience with v15.0rc1, the SSH server session would **never** timeout, leaving a "half open" SSH session hanging indefinitely (see https://github.com/turnkeylinux/tracker/issues/1092). As I noted over there:

> Whilst it's not implicitly a security risk (i.e. a hacker couldn't easily intercept the half-open connection) IMO it's still something of a risk. It relies on the user having their PC setup securely, e.g. requiring password on wake up.

TBH, I haven't explicitly tested it since he made this change (I'm still battling with intihooks), but my assumption is that unless there is some misunderstanding somewhere, this PR was to close https://github.com/turnkeylinux/tracker/issues/1092 (i.e. SSH connection left hanging half open indefinitely).

If you particularly want to keep `TCPKeepAlive no`, then I'm open to that, but IMO ensuring that the SSH connection is dropped after some period of time without client response is fairly important.

qrntz:
`ClientAlive...` will not close the connection if the route simply goes down (which happens when the machine is suspended/hibernated), but `TCPKeepAlive` will. `ServerAlive...` is a client setting and only applies to .ssh/config or /etc/ssh/ssh_config, we can't (nor should) enforce it server-side. As discussed in the SO question I linked, `TCPKeepAlive` used together with `ClientAlive...` is safe. Actually we should probably also set `ClientAliveInterval`, I thought it was set.

JedMeister:
Thanks @qrntz 

From what I can gather, `ClientAliveInterval` defaults to `0` and we're not adjusting that by default.

Dude4Linux:
@JedMeister & @qrntz - Thanks for the clarification. I missed the link to #1092 and only followed the link Anton provided to StackExchange.

After doing some more Googling, I'm as confused as ever, and judging by the number of different explanations and recommendations I've seen, a lot of other Admins are also confused.  It seems to me that we are really talking about two different concerns, idle user timeouts (@JedMeister) , and dropping "ghost" client connections (@qrntz).

@JedMeister - I tried to reproduce your results yesterday.  With the following in server's `sshd_config`:
```
# SSH hardening recommended by lynis
UsePrivilegeSeparation sandbox
X11Forwarding no
TCPKeepAlive no
AllowTcpForwarding no
ClientAliveInterval 150
ClientAliveCountMax 2
MaxAuthTries 2
MaxSessions 2
```
You were correct that no time out occurs after five minutes of user inactivity.  I would argue that this is expected behavior, i.e. keeping the connection alive as long as the `ssh` client responds to the `ClientAlive` packets from the server.  If you really want to disconnect after five minutes of user inactivity, try setting `ClientAliveCountMax 0` and `ClientAliveInterval 300`.  Seems odd, but it does work.

With the settings above, I logged into the server with two ssh sessions from my laptop, then suspended the laptop for the rest of the night.  This morning both ssh sessions had been dropped with this warning:
```
root@lxc ~# packet_write_wait: Connection to 192.168.1.62 port 22: Broken pipe
```
and in the server's auth.log
```
May 30 03:16:33 lxc sshd[19972]: Accepted password for root from 192.168.1.141 port 48466 ssh2
May 30 03:16:33 lxc sshd[19972]: pam_unix(sshd:session): session opened for user root by (uid=0)
May 30 03:24:41 lxc sshd[20212]: Accepted password for root from 192.168.1.141 port 48544 ssh2
May 30 03:24:41 lxc sshd[20212]: pam_unix(sshd:session): session opened for user root by (uid=0)
May 30 03:31:59 lxc sshd[19972]: Timeout, client not responding.
May 30 03:31:59 lxc sshd[19972]: pam_unix(sshd:session): session closed for user root
May 30 03:32:12 lxc sshd[20212]: Timeout, client not responding.
May 30 03:32:12 lxc sshd[20212]: pam_unix(sshd:session): session closed for user root
```
No half-open sessions were left behind.

In a discussion on [SuperUser](https://superuser.com/questions/1110690/why-do-inactive-ssh-sessions-terminate-with-write-failed-broken-pipe) Steffen Ullrich says:

> Who is closing the socket?
> 
> Nobody is closing the "socket", i.e. neither client nor server. The connection problem is usually caused by a stateful packet filteri in between the communication parties. Such packet filters are used in firewalls and also in the simple SoHo routers where they are needed to handle the states for NAT which is the typical way to provide internet access to multiple internal systems behind the same external IP address.
> 
> Such stateful packet filters create a state for each new connection and tears it down on connection close. This means SYN/FIN packets with TCP connections like ssh. These states takes up memory so a packet filter can only support a limited number of states at the same time. Therefore it will remove these states after some time of inactivity (i.e. no packets for this connection).
> 
> If there is no activity within a TCP connection no packets will be send. This is true for SSH too. After some time of inactivity the state in the packet filter will be removed. If any party of the connection then will send more packets after the inactivity there will be no state in the packet filter and a RST packet will be issued. This results in the "Broken Pipe".
> 
> The TCPKeepAlive option in SSH enables the TCP keepalive option which causes the TCP layer to regularly send empty packets (i.e. TCP packet with no data payload) in times of inactivity. This way the packet filter keeps the state open.
> 
I don't have a packet filter device between my laptop and my server so this is not an issue for me, however it might be for someone who is doing remote administration. +1 for setting `TCPKeepAlive yes`.

I think CISOfy may have jumped the gun on changing the recommendation for `TCPKeepAlive` from `yes` to `no`.  I also agree that both `ClientAliveCountMax` and `ClientAliveInterval` should be set, but I'm not sure of which values to choose.  In my previous job, the Senior Admin enforced a hard limit of five minutes idle time, which I found quite annoying.

JedMeister:
Thanks @Dude4Linux - After my own research, I too found it hard to get a conclusive answer. So I ended up posting a [question on security.stackexchange](https://security.stackexchange.com/questions/186655/openssh-server-are-there-any-security-implications-and-or-value-of-for-tcpkeep). The [answer](https://security.stackexchange.com/a/186660/81392) I got was pretty concise and seemed pretty straight forward and reasonable to me in light of all my other research.

You can read it yourself, but my takeaway was that we should definitely be setting `ClientAliveInterval` (and `ClientAliveCountMax` although apparently `0` is a legitimate value?!). `TCPKeepAlive` won't do any harm and isn't a security risk. However with `ClientAliveInterval` properly configured, it also doesn't have any real value (as it essentially serves the same purpose as `ClientAliveInterval` - albeit via a different mechanism). It's also worthy of note that the `TCPKeepAlive`mechanism may not be reliable in modern IT usage scenarios anyway.

So IMO we should configure `ClientAliveInterval` (and `ClientAliveCountMax`) appropriately and drop `TCPKeepAlive`. More because it has no value, than it being a security risk.

Dude4Linux:
@JedMeister - In Forest's answer that you linked to, I believe role of Client vs Server is actually the reverse:
> ClientAliveInterval - A message is sent periodically from the client to the server using the SSH protocol. The main benefit to using this is that a firewall cannot tell if it is a keepalive or a legitimate message. A TCP keepalive on the other hand may be blocked by a firewall.
> 
> ServerAliveInterval - Idem but sent from the server to the client.
Just another example of how much confusing information is floating around out there. :)

Nevertheless, I agree with your conclusion :
> So IMO we should configure ClientAliveInterval (and ClientAliveCountMax) appropriately and drop TCPKeepAlive. More because it has no value, than it being a security risk.

I suggest we set `ClientAliveCountMax 2`  as recommended by Lynis, and set `ClientAliveInterval 1h` to match `TCPKeepAlive`'s 2 hour default.

I'll take an action item to contact CISOfy and suggest they add a 'warning' to set `ClientAliveInterval` to a non-zero value if `TCPKeepAlive` is disabled.

PR #127, opened 2018-05-31
Replace the outdated .bashrc prompt code with a bash4 shell setting

qrntz:
I ran into this when investigating https://github.com/turnkeylinux/tracker/issues/1094 and noticed that the code is 7 years old and gets in the way sometimes.


JedMeister:
TBH, I'm not a massive fan of the ellipses (but perhaps that's just me and what I'm used to...). Regardless, it's makes for a MUCH nicer `.bashrc`! 

The more convention we use the better IMO and if we can reduce complexity, I'm all for it!

Merging...

PR #132, opened 2018-08-08
Removed some deprecated packages from nodejs conf script

OnGle:
Notes on package removals:

NSP's functionality has been merged into npm (see https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting)

Bower has been deprecated and referenced in a banner on the bower website (https://bower.io/) and presumably in other places, suggested alternatives include webpack and yarn, I know nothing about these packages and can't offer any suggestion moving forward regarding a bower replacement.

### DIFF
--- a/conf/nodejs
+++ b/conf/nodejs
@@ -1,5 +1,5 @@
 #!/bin/bash -ex
-GLOBALS='bower grunt grunt-cli nsp'
+GLOBALS='grunt grunt-cli'
 
 for package in $GLOBALS; do
 	npm install -g $package


JedMeister:
FWIW, I'm pretty sure that the Canvas app uses `yarn`, so perhaps that's the go?! Any thoughts @qrntz ?

qrntz:
Looks like bower [is superseded](https://snyk.io/blog/bower-is-dead/) by an arbitrary subset of npm, yarn and webpack. npm is not going anywhere since it's seen as standard by a lot of people/projects. npm as the standard option or yarn+webpack for projects that require it sounds good to me. npm and yarn kind of fill the same niche and for most uses npm is enough so I don't think it's justified to include it by default. npm has also been moving towards feature parity with yarn [from what I gather](https://medium.com/wemake-services/is-yarn-still-a-thing-3c6886410c83).

From out appliances canvas and gitlab use yarn. Perhaps we can use a common "fat-rails" template on top of the nodejs common code for those two apps to cut boilerplate.

JedMeister:
Awesome @qrntz , thanks for the added info. @OnGle had surmised something similar, but doesn't have much experience with NodeJS so was not as confident as you seem! :smile: 

> From out appliances canvas and gitlab use yarn. Perhaps we can use a common "fat-rails-nodejs" template on top of the nodejs common code for those two apps to cut boilerplate (as their requirements are somewhat similar).

Sounds like a good option to me mate.

PR #142, opened 2019-11-20
Updates for LAPP

OnGle:
Update postgres inithook to python3 ready for buster


JedMeister:
IMO, it'd be good to get rid of use of `os.system` - and standardise on `subprocess` (`.run`/`checkout` and `Popen` where required). But we'll leave it be for now...

PR #144, opened 2020-02-18
Add common mysql plan & update mysqlconf.py

OnGle:
- add common mysql plan
- update mysqlconf.py (mostly compatible, requires small modifications to use in inithooks)
    required for https://github.com/turnkeylinux/tracker/issues/1414


JedMeister:
Good work mate. Only thing I noticed was that the `webmin-mysql` package should probably also move to the new `mysql` plan.

And even though they're no longer relevant, the `lamp56` & `lamp72` plans should also be bought into line (actually perhaps they should just be removed altogether?!).

PR #148, opened 2020-03-31
Support for generating dhparams in turnkey-make-ssl-cert

JedMeister:
The first, and arguably biggest part of https://github.com/turnkeylinux/tracker/issues/1432

### DIFF
--- a/overlays/turnkey.d/sslcert/usr/local/bin/turnkey-make-ssl-cert
+++ b/overlays/turnkey.d/sslcert/usr/local/bin/turnkey-make-ssl-cert
@@ -335,9 +335,10 @@ if [[ "$GEN_DH" == "true" ]]; then
         msg="Diffie-Hellman parameter generation failed."
         error_exit "$msg OpenSSL output: " $TMPOUT
     elif [[ $DH_BITS -lt 2048 ]]; then
-        msg="Diffie-Hellman parameter bit size is $DH_BITS. Recommended"
-        msg="$msg minimum is $DH_BITS_REC; 4096 is best, but may take hours to"
-        warning "$msg generate."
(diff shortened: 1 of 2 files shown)


overlays/turnkey.d/sslcert/usr/local/bin/turnkey-make-ssl-cert
```diff
@@ -270,25 +299,57 @@ else
...
+        _time="hours, possibly days"
+    fi
+    info "Generating Diffie-Hellman parameters file - this may take $_time..."
+    if ! openssl dhparam -out $DHP $DH_BITS > $TMPOUT 2>&1; then
+        msg="Diffie-Hellman parameter generation failed."
+        error_exit "$msg OpenSSL output: " $TMPOUT
+    elif [[ $DH_BITS -lt 2048 ]]; then
+        msg="Diffie-Hellman parameter bit size is $DH_BITS. Recommended"
+        msg="$msg minimum is $DH_BITS_REC; 4096 is best, but may take hours to"
+        warning "$msg generate."
```
OnGle, line 340: This seems a little odd to me the way you're creating this I would've thought it cleaner to simply do

```bash
msg="Diffie-Hellman parameter bit size is $DH_BITS. Recommended
minimum is $DH_BITS_REC; 4096 is best, but may take hours to
generate"
```

(fwiw this will replace newlines with spaces, but in this case that appears to be intended)

JedMeister: Well spotted. That is certainly much nicer looking. FWIW, there are a couple more places that could be tidied, so I did them too! :smile:

PR #159, opened 2020-07-16
Web2py py3

JedMeister:
Add in Web2py support for python3 but leave as python2 by default.

See also: https://github.com/turnkeylinux-apps/web2py/pull/8

### DIFF
--- a/conf/web2py
+++ b/conf/web2py
@@ -8,6 +8,7 @@ dl() {
 SRC=/usr/local/src
 W2PPASS=turnkey
 W2PROOT=/var/www/web2py
+WSGI=$W2PROOT/wsgihandler.py
 
 # clone web2py
 cd /var/www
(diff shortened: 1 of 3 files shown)


conf/web2py
```diff
@@ -17,20 +18,28 @@ git clone --recursive --branch $VER https://github.com/web2py/web2py
...
-# Insert code just below initial imports
-sed -i "s|import os|import os\n\n$VIRTUALENV_ACTIVATE|" $W2PROOT/wsgihandler.py
+# Add code for entering virtualenv after wsgihandler is started
+#
+# Insert placeholders
+sed -i 's|import os|import os\n\nLINE_1\nLINE_2\nLINE_3|' $WSGI
+# Insert proper lines
+sed -i "s|LINE_1|activate_this = '$W2PROOT/venv/bin/activate_this.py'|" $WSGI
+sed -i "s|LINE_2|with open(activate_this, 'rb') as fob:|" $WSGI
+sed -i "s|LINE_3|    exec(compile(fob.read(), activate_this, 'exec'), dict(__file__=activate_this))|" $WSGI
```
OnGle, line 42: Why not use sed's `a` (append) for adding these lines rather than substitution?

JedMeister: Yeah, you're probably right. I did initially try to do it via an alternate way but I couldn't get it to work and IMO doing it the way I did makes it relatively readable. Regardless, I'm going to leave it as is for now...

PR #161, opened 2020-07-20
test for existence of 'a2enmod' rather than execute it blind

deutrino:
In the case that leftover apache debris exists in /etc for some reason, this change improves robustness of this particular script.

### DIFF
--- a/conf/adminer-apache
+++ b/conf/adminer-apache
@@ -1,8 +1,12 @@
 #!/bin/sh -e
 
-echo Listen 12322 >> /etc/apache2/ports.conf
-ln -s /etc/adminer/apache.conf /etc/apache2/sites-available/adminer.conf
-a2ensite adminer
+FILE=/etc/apache2/ports.conf
+BIN=/usr/sbin/a2ensite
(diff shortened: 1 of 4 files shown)


deutrino:
This was only happening due to user error (I failed to make a needed change to the Makefile of the appliance I'm developing, resulting in it erroneously executing apache-related conf scripts).

However if this is still a valuable property for these scripts to have, I fixed a few others to test for the existence of target files/executables - if desired I'll add that to this pull request, otherwise it can be closed.

deutrino:
added those checks just in case you want them.

I have a feeling that, with a bit more understanding how this works, it's better for those scripts to fail than to have the checks... but leaving my work here in case maintainers think different. Feel free to close if this PR is moot.

JedMeister:
Thanks for taking the time to provide this, but as the Apache related scripts you've modified should never be run unless the Apache config files exist.

I have no doubt that we could improve our docs to make it clearer, but if things are set up properly, the issues you hit should never occur.

As such, I'm going to close this. Thanks anyway...

PR #166, opened 2020-09-11
Some refactoring of gh_releases

JedMeister:
Primarily the changes are to support unauthenticated usage. But some other tweaks too.

Note: under some circumstances (e.g. multipage results) unathenticated usage may lead to errors and/or unreliable results. So is generally not recommended in production...

FWIW the core of this has been sitting in my local dev common for some time. This is just a commit of what I've been using locally for a while, with a few more tweaks...

### DIFF
--- a/overlays/github-latest-release/usr/local/bin/gh_releases
+++ b/overlays/github-latest-release/usr/local/bin/gh_releases
@@ -7,15 +7,18 @@ tmp_dir=/tmp/gh_releases
 rm -rf $tmp_dir
 mkdir -p $tmp_dir
 
-fatal() {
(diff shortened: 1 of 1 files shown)


OnGle (approved):
Sorry I completely missed this, absolutely my fault for being slack with my notifications.

Yeah I don't have any issues with this explicitly, it makes sense to me to support this. As you note occasionally it *can* lead to unexpected results or errors with multi-page results and the query limit gets hit, perhaps long-term that could be addressed by somehow triggering a hard build failure if it does happen?

JedMeister:
Thanks man. I'll merge this now then, but I like your idea of an error if the query limit gets hit.

PR #170, opened 2020-11-11
Refactor secupdates (cronapt) to never remove apps

JedMeister:
Closes https://github.com/turnkeylinux/tracker/issues/1536

Use 'upgrade' with APT::Get::Upgrade-Allow-New=true, rather than
'dist-upgrade' (allows new packages to be installed, but never removes
any).

Also improved layout for easier reading.

### DIFF
--- a/conf/turnkey.d/cronapt
+++ b/conf/turnkey.d/cronapt
@@ -1,8 +1,12 @@
-#!/bin/sh -e
+#!/bin/bash -e
+
+# Note that a Confconsole plugin that allows the user to select between these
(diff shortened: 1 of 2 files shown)


JedMeister:
@OnGle

OnGle (approved):
Looks good to me :+1:

(p.s. extra :+1: for the typing)

JedMeister:
Oops, I just noticed that the path of the files was wrong in the Confconsole plugin. I've just fixed that...

PR #175, opened 2021-02-03
Confconsole perf and info schema install

JedMeister:
Implements the Performance and information schemas, tools, etc; inc SysSchema to MariaDB.

Closes https://github.com/turnkeylinux/tracker/issues/1429


OnGle (approved):
Looks good to me

PR #178, opened 2021-02-09
Include composer wrapper (turnkey-composer) in common

JedMeister:
This will now be included in all LAMP/LAPP based appliances, plus any appliance that includes `composer.mk` or the `composer` common overlay.

Part of the solution to https://github.com/turnkeylinux/tracker/issues/1539.


OnGle (approved):
Awesome!

PR #179, opened 2021-02-09
Rather than comment out eth1; instead configure it as 'hotplug'.

JedMeister:
Closes https://github.com/turnkeylinux/tracker/issues/1492.


OnGle (approved):
:+1:

PR #180, opened 2021-02-09
Install qemu-guest-agent by default.

JedMeister:
I've ummmed and arrrred about this one for a while, but seeing as it's only ~700kb when installed, I figured what the help, let's just do it...

Closes https://github.com/turnkeylinux/tracker/issues/1356.


JedMeister:
Oops, it looks like I accidentally merged this without waiting for your review @OnGle ...

PR #181, opened 2021-02-10
Remove composer from default LAMP and LAPP installs

JedMeister:
Remove from common `lamp.mk` & `lapp.mk` makefiles. We'll re add it to the `Makefile` of specific apps that use it, or we deem relevant (e.g. LAMP and LAPP appliances).

Part of https://github.com/turnkeylinux/tracker/issues/1563


OnGle (approved):
:+1:

PR #184, opened 2021-03-30
Add in turnkey-artisan wrapper script and makefile for convenience.

JedMeister:
This one is currently only used in Bagisto, but `artisan` is a Laravel component/command, so will likely apply to Laravel and other Laravel based apps.

I have included a `artisan.mk` file too, although strictly speaking it isn't really required (and perhaps I shouldn't have). It could just be included  via the `COMMON_OVERLAYS += artisan` (in the appliance `Makefile`) but I sort of prefer including the `mk` file...

Or actually, perhaps the make file should be `laravel.mk`and include `composer.mk`?

I'm happy to make changes, or to merge as is... Your thoughts @OnGle?


OnGle (changes requested):
In my opinion if artisan is relevant for all laravel apps, that the Makefile should be `laravel.mk` for further extension if necessary. Be interested to hear your thoughts on this

JedMeister:
> the Makefile should be `laravel.mk` for further extension if necessary.

Great suggestion! That makes tons of sense...

I'll update to use that. Also, as Laravel apps will also leverage Composer, it should depend on `composer.mk` too.

OnGle (approved):

PR #187, opened 2021-04-07
Include gh_releases script in all builds

JedMeister:
Note that it is always removed towards the end of the build anyway, so no problem if it's not needed.

@OnGle


JedMeister:
Ideally I should wait for @ongle before I merge this. OTOH, this is pretty straight forward. Also the Etherpad update (that I really want to build right now) requires this now. I guess that I could add it to Etherpad itself and merge that without a review instead, but that seems no better, in fact worse...

So...

PR #188, opened 2021-04-08
build time `service`/`systemctl` overlay tweaks

JedMeister:
The changes here were initially and primarily to support the v16.1 [Odoo](https://github.com/turnkeylinux-apps/odoo) update ([see here](https://github.com/turnkeylinux-apps/odoo/pull/18)).

I have also done some refactoring and tidying up.

This PR is also aimed at allowing the `systemctl` wrapper to work with [Ghost](https://github.com/turnkeylinux-apps/ghost) (and replace the existing overlaid [Ghost specific `systemctl` wrapper](https://github.com/turnkeylinux-apps/ghost/blob/master/overlay/usr/local/bin/systemctl)). Thus it is part of https://github.com/turnkeylinux/tracker/issues/1504.

Moving forward, we can tweak these scripts to support appliances on an "as needs be" basis.

As much as possible, IMO it would be good to keep the `systemctl` wrapper relatively generic, with the `service` wrapper to deal with specific idiosyncrasies of individual apps.

FYI, if the `/usr/sbin/service` command detects that systemd isn't running (as in a chroot) then it defaults to using `/etc/init.d` scripts. Any daemons that ship with `init.d` scripts (that work in a chroot) should continue to "just work", whilst we can make specific tweaks in our `service` wrapper as needed for other specific use cases.

### DIFF
--- a/conf/turnkey.d/shellinabox
+++ b/conf/turnkey.d/shellinabox
@@ -11,9 +11,9 @@ EOF
 # TurnKey is now providing a shellinabox systemd service file.
 # First disable the current service (which is generated from the init.d script
 # via systemd-sysv integration):
-systemctl disable shellinabox
+/lib/systemd/systemd-sysv-install disable shellinabox || true
 # Then explictly enable the TurnKey provided service file:
-systemctl enable /usr/lib/systemd/system/shellinabox.service
+systemctl enable shellinabox.service || true
 
 # Stunnel TLS/SSL support for shellinabox (aka Webshell) is now configured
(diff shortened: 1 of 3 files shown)


OnGle (approved):
I don't really like the app specific changes to `systemctl` but I also don't see an easy alternative for a point release knowing that ghost will try to use it directly. IMO this non-ideal fix is a symptom of our non-ideal handling of services and as such makes sense to implement until we want to overhaul this weird service-in-chroot wrapping setup we have, (which wont happen until ATLEAST 17.0)

JedMeister:
Yeah fair comments...

PR #201, opened 2021-12-03
TEMP - build 17.x on v16.x TKLDev.

JedMeister:
Much of this can be reverted once we are ready for v17.0 stable, but is currently needed to build 17.0rc on v16.x TKLDev.

I'm merging now, but please check it out @OnGle.


JedMeister:
This was a mistake and I ended up force pushing to clear it from the history (not ideal I know...). Please see #202 for the proper version of this...

PR #205, opened 2022-01-18
Apache fix up

JedMeister:
This really should have been part of https://github.com/turnkeylinux/common/pull/204.

Primarily it enables the Apache headers module (i.e. runs `a2enmod headers`). This is required for HSTS support (which is now included by default in ssl.conf).

I have refactored things a bit so that now instead of adhoc including specific Apache related conf scripts and overlay, there is a dedicated `apache.mk` file.


JedMeister:
@OnGle

OnGle (approved):
lgtm :+1:

PR #206, opened 2022-01-23
Kernel hardening

JedMeister:
Install `hardening-runtime` (kernel hardening) package (new in Buster), plus adjust default kernel TurnKey hardening setting filename (to avoid overwriting the one included in `hardening-runtime`).

@OnGle @qq7


OnGle (commented):
lgtm, if you can confirm there aren't any conflicts that would lead to an effective regression in hardening then consider it approved.

JedMeister:
@OnGle - as discussed offline, I am rethinking this addition.

As per https://github.com/turnkeylinux/tracker/issues/1703 - instead of including `hardening-runtime` by default for v17.0, I'm now thinking that we should just move our hardening config for now and leave inclusion of `hardening-runtime` for a later date.

The rationale is whilst I'm fairly confident that this should "just work", there is a risk that it may cause issues for some appliances, in some environments and some scenarios. At this late stage of the v17.0 dev cycle, the (potential) cost/benefit just doesn't really stack up.

JedMeister:
FWIW, I just cherry picked https://github.com/turnkeylinux/common/pull/206/commits/a9d3eec74949898ec6776f438fc01b648b1e4521 into master and pushed. So the existing TurnKey hardening is as it was, but now renamed so no future risk of clash with `hardening-runtime`.

PR #207, opened 2022-02-01
Nodejs v17.0

qq7:


JedMeister:
I haven't tested, but it looks good to me man.

My only thought is that I note the inclusion of a sysvinit script for pm2. IMO we should move to a systemd service file instead (obviously what's there now should work due to systemd backwards compatibility with old sysvinit, but better to just use the system default IMO).

From what I can gather pm2 provides facility to generate systemd service files?! Here's some relevant looking links I found doing a quick search:
- https://pm2.keymetrics.io/docs/usage/startup/
- https://pm2.io/docs/runtime/guide/startup-hook/
- https://www.tecmint.com/enable-pm2-to-auto-start-node-js-app/

Perhaps have a quick look yourself and see what you think? If it's easy enough to update to using that scenario, let's do that for v17.0. If it doesn't "just work", then let me know and we can just leave it as is (i.e. I'll merge this PR as it is) and we'll open a "feature" issue for future investigation.

qq7:
@JedMeister 

Done.

OnGle (approved):

PR #208, opened 2022-02-02
Update pgsqlconf.py

qq7:


JedMeister:
Great stuff. Nice work on the general code cleanup too!

PR #212, opened 2022-03-28
Last minute

JedMeister:
Some last minute tweaks:
- move Secupdates_adv_conf.py confconsole plugin to confconsole source - see https://github.com/turnkeylinux/confconsole/pull/57
- some minor tidying for zz-ssl-ciphers conf script


conf/turnkey.d/zz-ssl-ciphers
```diff
@@ -47,14 +47,15 @@ fi
...
     sed -i "/CipherString/ s|ZZ_SSL_CIPHERS|$SECURE_CIPHER_LIST|" $CONF
     lighty-enable-mod ssl
 fi
 
 # Tomcat
-CONF="/etc/tomcat7/server.xml"
-if [ -f $CONF ]; then
+# As of v15.x TKL uses Debian Tomcat default ciphers
+CONF="/etc/tomcat*/server.xml"
+if [ -f "$CONF" ]; then
```
OnGle, line 58: Will this expand? Afaik `*` only expands when un-escaped...

JedMeister: Yep you're right. I'll fix it right away...

JedMeister: Good find. On face value it looked good to me, but you are right. I just tested and it checks for the literal string `/etc/tomcat*/server.xml`.

PR #218, opened 2022-04-22
Add dynamic sury pinning

OnGle:
adds variable `PHP_EXTRA_PINS` which can be set with space seperated
package names. These will be pinned to 550 from sury.


JedMeister:
Nice work man! :)

PR #225, opened 2022-09-16
also build 'php-common' dummy package for use with 3rd party PHP.

JedMeister:
Hack to workaround PHP7.3 install issues for v17.x.

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -18,9 +18,9 @@ fatal() {
 }
 
 build_dummy_pkg() {
-    PKG=$1
+    declare PKG=$1
     mkdir -p /tmp/$PKG/DEBIAN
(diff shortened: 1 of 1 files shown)


conf/bootstrap_apt
```diff
@@ -17,6 +17,30 @@ fatal() {
     exit 1
 }
 
+build_dummy_pkg() {
+    PKG=$1
+    mkdir -p /tmp/$PKG/DEBIAN
+    PKG_V="2:${PHP_VERSION}"
```
OnGle, line 23: PKG and PKG_V could be locals here `local PKG=$1 PKG_VER=...` or `declare PKG=$1 PKG_VER=...`

JedMeister: Great point! Thanks man. I'll apply that change then merge this so we don't hold up Mattie.

PR #226, opened 2022-09-16
Revert previous changes and actually fix the issue

JedMeister:
It turns out that I didn't understand the issue properly. Unlike the other php-PKG_NAME packages that have a phpX.Y-PKG_NAME counterpart, the php-common package, actually includes some important scripts. Namely the phpenmod and phpdismod commands. So a real php-common package needs to be installed from somewhere, or php module package installs will fail (when they call 'phpenmod').

Installing the default Debian one, pulls in other default Debian PHP packages so that's not an option. But the sury.org php-common installs cleanly. It also does not pull in the latest phpX.Y-common as other sury.org php-PKG_NAME packages do (e.g. php-mysql will pull in the latest phpX.Y-mysql).

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -17,30 +17,6 @@ fatal() {
     exit 1
 }
 
-build_dummy_pkg() {
-    declare PKG=$1
-    mkdir -p /tmp/$PKG/DEBIAN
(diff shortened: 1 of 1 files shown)


JedMeister:
I'm going to merge this straight in.

PR #228, opened 2022-09-29
Update common for faveo-helpdesk appliance; Depends on 

Christie018:
Updating common plan to include lamp plan for php version 7.3 specifically for faveo-helodesk appliance. Changes made to common/conf/bootstrap_apt ensure pin priority is set correctly to install php packages from sury.org repository instead of debian repo.

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -117,10 +117,6 @@ deb [signed-by=$keyfile] https://packages.sury.org/php/ $CODENAME main
 EOF
 
     cat > /etc/apt/preferences.d/php-sury.pref <<EOF
-Package: *
-Pin: origin packages.sury.org
-Pin-Priority: 10
-
 Package: php${PHP_VERSION}-*
 Pin: origin packages.sury.org
 Pin-Priority: 550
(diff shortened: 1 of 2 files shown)


conf/bootstrap_apt
```diff
@@ -149,6 +145,10 @@ Package: php-common
 Pin: origin packages.sury.org
 Pin-Priority: 550
 
+Package: *
```
JedMeister, line 148: This bit needs to go at the top. It pins the priority for all packages from the sury repo to `10` (meaning that can't be installed, unless you explicitly install them). Then the packages that we do want to install are pinned to `550`, meaning slightly higher priority than the default Debian packages.

If you needed to do this to get it to install cleanly, then there must be something that it's installing that the default pinning is blocking. I suspect that there is a specific dependency that won't install cleanly.

I'm happy to have a look myself at some point, but I'm pretty under the pump, so likely won't be this week.

JedMeister (commented):
Looks ok Mattie, although I don't like the pinning changes you've made. Please see my comment for details, but basically it looks like your change was to work around an issue, but the result is that it defeats the point of our pinning.

conf/bootstrap_apt
Christie018: I went through and changed the code back, but I will have to go through and do more testing to get successful build with common still being correct.

Christie018:
I reviewed the initial PR comments and changed the code to reflect the suggested changes. The new PR's code changes have been retested with faveo-helpdesk appliance to ensure functionality.

OnGle (approved):
lgtm :)

JedMeister:
:tada:

PR #194, opened 2021-07-14
improve IPv6 support

richard-vd:
* install package webmin-firewall6 by default
* listen on port 12320 (shellinabox) and 12321 (webmin) both for IPv4 and IPv6

closes https://github.com/turnkeylinux/tracker/issues/1658

PR #229, opened 2022-09-29
improve IPv6 support

richard-vd:
In my previous PR (#194) I forgot to take the `webmin-conf-stunnel` script into account. This adds the missing part.


JedMeister:
Excellent, thanks. I'll merge straight away. Unfortunately none of the appliances that I have already published will get this. But all ones from here on in will.

PR #231, opened 2022-11-08
Remove nginx from nodejs plan, move to new nodejs-nginx plan.

JedMeister:
Closes https://github.com/turnkeylinux/tracker/issues/1772

Also requires:

- https://github.com/turnkeylinux-apps/etherpad/pull/18
- https://github.com/turnkeylinux-apps/ghost/pull/16
- https://github.com/turnkeylinux-apps/mongodb/pull/20
- https://github.com/turnkeylinux-apps/nodejs/pull/15
- https://github.com/turnkeylinux-apps/redis/pull/8

PR #233, opened 2022-11-09
Create new 'nginx-nodejs.mk' file and remove nginx from 'nodejs.mk'

JedMeister:
I thought I was done with Nginx/NodeJS tweaks, but @OnGle noted that `mk/turnkey/nodejs.mk` also included Nginx components. So to be consistent with the previous PR (https://github.com/turnkeylinux/common/pull/231) I'm updating the makefiles too. `mk/turnkey/nodejs-tklwebcp.mk` always requires Nginx so that is updated too.

I've also noticed an oversight in my previous updates. The `overlays/nodejs` only includes the Nginx reverse proxy config, so should be part of the `mk/turnkey/nginx-nodejs.mk`, not `mk/turnkey/nodejs.mk`. Also the overlay should be renamed ` overlays/nodejs-nginx` to be more explicit.

That does leave the `mk/turnkey/nodejs.mk` file only including `conf/nodejs` - but let's run with that for now. I.e. for now `include $(FAB_PATH)/common/mk/turnkey/nodejs.mk` in an appliance `Makefile` is functional equivalent to `COMMON_CONF = nodejs`.

So this is (hopefully) the final part of https://github.com/turnkeylinux/tracker/issues/1772.

Specific app related required updates:

- https://github.com/turnkeylinux-apps/etherpad/pull/19
- https://github.com/turnkeylinux-apps/ghost/pull/17

### Appliances that are (potentially) affected - but don't need to be changed:

_**Doesn't**_ need Nginx:

- Bagisto (`Makefile: COMMON_CONF = nodejs`)
- Canvas (`Makefile: include $(FAB_PATH)/common/mk/turnkey/nodejs.mk`)
- Invoice Ninja (`Makefile: include $(FAB_PATH)/common/mk/turnkey/nodejs.mk`)
- Joomla4 (`Makefile: COMMON_CONF = nodejs`)

_**Does**_ need Nginx but will get it via updated `nodejs-tklwebcp.mk`:

- CouchDB (`Makefile: include $(FAB_PATH)/common/mk/turnkey/nodejs-tklwebcp.mk`)
- MongoDB (`Makefile: include $(FAB_PATH)/common/mk/turnkey/nodejs-tklwebcp.mk`)
- NodeJS (`Makefile:include $(FAB_PATH)/common/mk/turnkey/nodejs-tklwebcp.mk`)
- Redis (`Makefile: include $(FAB_PATH)/common/mk/turnkey/nodejs-tklwebcp.mk`)


mk/turnkey/nodejs-tklwebcp.mk
```diff
@@ -1,4 +1,4 @@
 COMMON_OVERLAYS += tkl-webcp nodejs-pm2-tklwebcp
 COMMON_CONF += tkl-webcp nodejs-pm2-tklwebcp
 
-include $(FAB_PATH)/common/mk/turnkey/nodejs.mk
+include $(FAB_PATH)/common/mk/turnkey/nginx-nodejs.mk
```
OnGle, line 4: Include should go before `COMMON_OVERLAYS` and `COMMON_CONF` as `nodejs-pm2-tklwebcp` requires `nodejs`

mk/turnkey/nginx-nodejs.mk
```diff
@@ -0,0 +1,4 @@
+COMMON_OVERLAYS += nodejs-nginx
+
+include $(FAB_PATH)/common/mk/turnkey/nodejs.mk
+include $(FAB_PATH)/common/mk/turnkey/nginx.mk
```
OnGle, line 4: Includes should probably go above the overlay here. It won't cause issues currently, but this means any overlays included from these 2 files will be applies AFTER `nodejs-nginx` which in the future may cause unexpected results.

OnGle (changes requested):

mk/turnkey/nodejs-tklwebcp.mk
JedMeister: Hmm, I wonder if this has caused minor issues and.or weirdness previously. Anyway, I'll put the include line first.

OnGle: I thought it was similar to the cause of the redis issue, but it actually IS the cause of that issue.

mk/turnkey/nginx-nodejs.mk
JedMeister: Hmm, I'm not 100% sure on this.

Looking at the scripts, `overlay/nodejs-nginx` only includes (node specific) nginx config (so doesn't matter when it's applied). `nodejs.mk` is the one that only includes `conf/nodejs` (which just sets up node and the `node` user) and `nginx.mk` includes nginx conf and overlay (neither of which are impacted either way by node specific stuff).

So long as overlays don't overwrite one another (which we possibly should have some text/check for? Perhaps a warning when overwriting any files ever, but an error if overlayed files are overwritten by other overlays? Not sure how realistic that might be though...) 

Anyway, having said all that, I don't have any real objection on changing it, but I don't think it's really required.

OnGle (approved):
lgtm!

mk/turnkey/nginx-nodejs.mk
OnGle: I agree, however it seems to me un-intuitive that files depending on both `nginx` and `nodejs` to be written to the filesystem prior to the specific overlays for `nginx` and `nodejs` respectively.

And it seems possible to me that naive changes to one of these 3 overlays breaks the assumption that they don't interact. Re-ordering ensures that the most-specific overlay wins, rather than the least specific, which (at least I) would not expect had I made those naive changes.

-----

TL;DR agree with your case, but arrive at a different conclusion.

PR #234, opened 2022-11-17
Initial BashLib implementation

JedMeister:
This is an initial implementation of "BashLib" as has been threatened (and desired) for some time.

It has been developed alongside v17.x update of CouchDB and Etherpad so only includes functions specific to (and now required by) them so there is lots of room for extension. Also, whilst I'm pretty comfortable with bash, there are so many different ways to do things in bash, I'm sure that the current code could be further generalised and/or improved.

The last commit (new `tkl-webcop.mk`) is unrelated to bashlib and probably should be in a separate PR, but it is required for couchdb and I wanted to finalise this too: https://github.com/turnkeylinux-apps/couchdb/pull/17


OnGle (approved):
lgtm :+1:

OnGle (approved):
lgtm! :)

PR #238, opened 2022-12-15
Support returning output from MySQL (i.e. support 'SELECT').

JedMeister:
Minor tweak to allow getting output from MySQL - e.g. return data from a `SELECT ...` query.


OnGle (approved):
lgtm

PR #241, opened 2023-02-07
Use full path to runuser so that it works when run via cron.

JedMeister:
When run as cron jobs, artisan commands run via `www-data` user will fail (`www-data` has no shell). So a better way to go is to use turnkey-artisan wrapper script. However, in that scenario, the full path to sbin commands is required.


OnGle (approved):
lgtm

PR #242, opened 2023-02-21
Bitkey revival

JedMeister:
It's only early days, but I thought I might have a look at reviving BitKey (https://github.com/bitkey/bitkey).

This PR will need further work, but there has been some progress.


JedMeister:
Whilst this is not complete (i.e. bitkey still doesn't build) I figure that this shouldn't cause any regressions and we may as well merge this. Again it's a case of using the transition as an opportunity to consolidate code that has already been developed.

OnGle (approved):
lgtm

JedMeister:
As per other common PRs I'll re-target this against `18.x-dev` and merge now.

PR #244, opened 2023-03-07
Backport nonfree

OnGle:
Add new CONFVAR for using non-free backports


Christie018 (approved):
Changes to add variable for including backports look good to me. I would suggest including something along the lines of "required by OTRS PR#20" so that devs know the two are related. I see you already have something like this for the OTRS PR so maybe just something in the comments will do 👍🏾

OnGle:
Good call, required by https://github.com/turnkeylinux-apps/otrs/pull/20

JedMeister:
LGTM! Merging.

PR #247, opened 2023-05-24
Added FTP option using pure-ftpd (for turnkey-fileserver appliance).

anandphulwani:
Created iso on TKLDev, tested and working.


JedMeister:
Hi @anandphulwani - Thanks for your contribution and apologies for such a huge delay responding to you. FWIW we're deep in dev mode, working on v18.0 (RCs of at least Core and TKLDev should hopefully be released next week, with the first batch of stable apps released soon after Debian releases their first Bookworm point release (i.e. v12.1 - due in abut a month).

There have been similar suggestions in the past but I've resisted because OpenSSH provides SFTP out of the box and vanilla FTP is insecure and rarely used these days. Although I do note that including this would add FTPS support (which currently .

Were you aware of the baked in SFTP support? With that in mind do you think that it's worth adding FTPS support too?

I see you've chosen to included `pure-ftpd` (which I am unfamiliar with) - is there a particular reason you chose that one above the other options? (Debian appears to have a few packaged; namely [pure-ftpd](https://packages.debian.org/bookworm/pure-ftpd) (as suggested/used by you), [`proftpd`](https://packages.debian.org/bookworm/proftpd-core), [`vsftpd`](https://packages.debian.org/bookworm/vsftpd) and/or [`ftpd-ssl`](https://packages.debian.org/bookworm/ftpd-ssl).

Also, when adding additional services, where possible, to make it easier for newbs, we try to also provide Webmin support.  I note that Webmin only has a core module for `proftpd` (it also has a `wuftp` module, but that isn't available via Debian). There is a third party vsftpd module which we could look at. There do not appear to Webmin modules for pure-ftpd or ftpd-ssl. It's not the end of the world if there isn't a relevant webmin module, but it is good where possible.

If size weren't a consideration, I'd be inclined to suggest using proftpd (with it's accompanying webmin module, and perhaps even it's clamav module too) instead. But I note that ProFTPd is significantly larger; not counting deps (but including it's crypto module, for FTPS support) is ~7MB vs `pure-ftpd` = ~1MB! So perhaps that's why you chose it? Although if you are purely going for small install size, did you consider [`vsftpd`](https://packages.debian.org/bookworm/vsftpd) or [`ftpd-ssl`](https://packages.debian.org/bookworm/ftpd-ssl) - ~340KB and ~160KB respectively?

Bottom line, I'm open to considering this, but would like to understand your use case, software selection and thoughts regarding the MVP (minimum viable product) of your suggestion/request/contribution.

anandphulwani:
Hi @JedMeister, no issues with the delay, i had myself prolonged a PR for good 3 years because of being occupied, first of all i would like to mention i am big fan of turnkey products and have been following/using for the past good 12-13 years and congrats for the work you are doing on the new version.

On the PR, i wish to clarify that i had been waiting for FTP support, and a lot had been cooking on the forums for past 10-12 years, including many times developers initiating development for this, but nothing came out, so in the end i thought of finishing it up myself. To sum up there is definitely a requirement of this among the users.

Now to clarify all the doubts, i would like to discuss my use case, i have been a very big fan of backup/imaging utilities like Acronis TrueImage, a lot of these utils support FTP/FTPS only, in my case i have been using Acronis, and it doesn't support SFTP, i deployed the original appliance and checked it as well by using the credentials on FTP tab, but it doesn't support.

So continuing my use case, i tried deploying vsftpd natively, but Acronis as a product wasn't behaving properly, it was taking too much time, i checked the ftp server logs and found out that Acronis makes multiple connections and makes a lot of parallel requests simultaneously and hangs the server, since i had to take backup now, and i could not delay, i took backup on one of my hosting providers, and everything went smoothly, then i checked cpanel and cpanel docs, found out they are using pure-ftpd, so i chose that as an option. Since a company like cpanel decision to use pure-ftpd will be a thought through decision.

I also looked for the webmin module for pure-ftpd and found out a community module from 2006, i tried integrating it but it doesn't go with the newer pure-ftpd version, and i knew that proftpd has a working webmin module, but if i remember correctly i found it a little complicated to set it up, though i am sure i didn't consider size requirements, also i will have to confirm this once but i think every user created normally as a linux user, in the webmin will have FTP/FTPs access to its own directory automatically.

In the end, i integrated pure-ftpd to the appliance and tested everything, taking backups on it, recovering from it, it works like a charm, so for switching to proftpd, i thought of the quote, which a wise man said, "don't tamper with something, which works perfectly". I am open to switch and integrate proftpd at a later stage, when i have time to test and check it thoroughly.

BTW, i have idea of a couple of fresh appliance which i am going to create for myself, will you be willing to add it in your library if it made sense to you.

JedMeister:
Ok cool. Thanks for providing the additional context and info. I'm happy with that.

So turning to your code specifically there are a couple of things of note. I'll do that inline (as a "code review") to make it a little easier to follow.

conf/ftp
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh -e
+
+# Copying cert.pem and dhparams.pem to pure-ftpd.pem and pure-ftpd-dhparams.pem.
+cp /etc/ssl/private/cert.pem /etc/ssl/private/pure-ftpd.pem
```
JedMeister, line 4: Rather than copy the default cert, I'd be inclined to either reconfigure pure-ftpd to use the default (i.e. `/etc/ssl/private/cert.pem`) or, if that's not easily done, just create symlinks (rather than copies).

My rationale for that is that by default these certs should be (re)generated at first boot, but the default certs are already regenerated. So if you leverage the defaults (either via using them directly or via symlinks) then you get their regeneration for free.

Also, it means that if a user configures Let's Encrypt, then FTPS will use the proper TLS/SSL certs by default too (and again for free).

conf/ftp
```diff
@@ -0,0 +1,27 @@
...
+cp /etc/ssl/private/dhparams.pem /etc/ssl/private/pure-ftpd-dhparams.pem
```
JedMeister, line 5: As per line above...

conf/ftp
```diff
@@ -0,0 +1,27 @@
...
+sed -i "s|^root$|# root|g"  /etc/ftpusers
+
+# Configure settings for pure-ftpd
+echo yes > /etc/pure-ftpd/conf/ChrootEveryone
+echo yes > /etc/pure-ftpd/conf/Daemonize
+echo no > /etc/pure-ftpd/conf/NoAnonymous
+echo 0 > /etc/pure-ftpd/conf/MinUID
+echo yes > /etc/pure-ftpd/conf/AllowUserFXP
+echo 1 > /etc/pure-ftpd/conf/TLS
+echo HIGH:MEDIUM:+TLSv1:\!SSLv2:\!SSLv3 > /etc/pure-ftpd/conf/TLSCipherSuite
```
JedMeister, line 17: Whilst this looks pretty secure, we have a [common conf script](https://github.com/turnkeylinux/common/blob/master/conf/turnkey.d/zz-ssl-ciphers) so we only need to update ciphers in one single place and then everything else gets them. I'd prefer it if the pure-ftpd config leveraged this same thing - that will reduce the risk that it might get overlooked in a future update.

JedMeister (changes requested):
All in all looks good. I just have a couple of things I've highlighted. I'm happy to discuss further if you disagree with my suggestions/requests.

conf/ftp
anandphulwani: I have reconfigured pure-ftpd to use the default cert located at `/etc/ssl/private/cert.pem`, but to do that i had to change the mechanism to use the `.conf` file instead, for which i had to make the following changes in `\etc\init.d\pure-ftpd` script.
- Disabled the `pure-ftpd-wrapper` which reads the config parameters files created in the `/etc/pure-ftpd/conf/` folder, for e.g. `/etc/pure-ftpd/conf/AllowUserFXP` and converts them into parameters being passed into pure-ftpd executable.
- Enabled the pure-ftpd executable to read the parameters from `.conf` file instead (On `start` and `restart` actions).

I am ready with the changes, and can update repo/PR, if you think its fine.

conf/ftp
```diff
@@ -0,0 +1,20 @@
...
+
+# Remove root from users that cannot login into the system via FTP
+sed -i "s|^root$|# root|g"  /etc/ftpusers
+
+# Configure settings for pure-ftpd
+echo yes > /etc/pure-ftpd/conf/ChrootEveryone
+echo yes > /etc/pure-ftpd/conf/Daemonize
+echo no > /etc/pure-ftpd/conf/NoAnonymous
+echo 0 > /etc/pure-ftpd/conf/MinUID
+echo yes > /etc/pure-ftpd/conf/AllowUserFXP
```
anandphulwani, line 15: These all parameters are being shifted to be configured from the `.conf` instead.

conf/ftp
anandphulwani: Have added the following block to `zz-ssl-ciphers` file

```
# Pure-FTPd
CONF="/etc/pure-ftpd/pure-ftpd.conf"
if [ -f "$CONF" ]; then
    sed -i "/^#\{0,1\} \{0,\}TLS    / s|.*TLS\(  *\)[0-9][0-9]*|TLS\11|" $CONF
    sed -i "/^#\{0,1\} \{0,\}TLSCipherSuite    / s|.*TLSCipherSuite\(  *\)..*|TLSCipherSuite\1HIGH:MEDIUM:+TLSv1:\!SSLv2:\!SSLv3|" $CONF
    
    sed -i "/^#\{0,1\} \{0,\}CertFile    / s|.*CertFile\(  *\)..*|CertFile\1$CERTFILE|" $CONF
fi
```

This changes values in `.conf` files, in the following way 
- `TLS` is changed to `1` (which accepts secure FTP as well as insecure plain FTP)
- `TLSCipherSuite` is changed to `HIGH:MEDIUM:+TLSv1:!SSLv2:!SSLv3`
- `CertFile` is uncomment'ed to enable it, and `$CERTFILE` path is used (from the variable's value declared above).

conf/ftp
JedMeister: > to do that i had to change the mechanism to use the .conf file instead, for which i had to make the following changes in \etc\init.d\pure-ftpd script.

Hmm, did you test using a symlink? Perhaps that might "just work"?

Also, by default the cert.pem and cert.key files are only readable by root, so be aware of that.

conf/ftp
JedMeister: Ok, that seems pretty good. Although TLS1 & TLS1.1 should also be disabled.

conf/ftp
anandphulwani: I am sure symlink would work out of the box, I just wanted to use `.conf` file method instead. 

Regd the readable by root, I already have implemented and tested the `.conf` file which is using the default cert file and it has no problem reading it.

conf/ftp
anandphulwani: I will check what can be done on disabling  TLS1 & TLS1. 1.

conf/ftp
anandphulwani: @JedMeister 

- I have tested using the `linking method`, creating symbolic link of cert.pem and dhparams.pem, it is working without any issues.
- I have also tested using the `.conf file method`, that is also working fine.

Please do suggest which method do you want me to implement?

conf/ftp
anandphulwani: I have successfully disabled TLS1 and TLS1.1 as well, and created options for both the methods `linking method` as well as `.conf file method` that i have mentioned earlier. I have used the testssl.sh to confirm (PFA attachment for the same).


testssl sh

conf/ftp
anandphulwani: @JedMeister Reply awaited.

conf/ftp
anandphulwani: @JedMeister Reply awaited.

JedMeister: I don't see any updates? Have you pushed them?

anandphulwani: @JedMeister Not yet, as i wanted your advise over it as i mentioned here
https://github.com/turnkeylinux/common/pull/247#discussion_r1282006596
> @JedMeister
> 
>     I have tested using the `linking method`, creating symbolic link of cert.pem and dhparams.pem, it is working without any issues.
> 
>     I have also tested using the `.conf file method`, that is also working fine.
> 
> 
> Please do suggest which method do you want me to implement?

conf/ftp
JedMeister: Thanks for the bump @anandphulwani - apologies on my slow response...

Let's just go with the symlink.

conf/ftp
JedMeister: My apologies. Responded above.

JedMeister:
Apologies again on slow response.

FWIW here's a bit of a status report relative to your work here:

I am currently working on the final touches before releasing tkldev and core (and hopefully lamp too) as v18.0 stable (should happen within the next week - fingers crossed). Once that is done, I plan to start publishing the apps that have already got updated code (mostly LAMP based apps). Only after we've got the low hanging fruit done there will be move on the fileserver appliance (and it's dependants; torrentserver & mediaserver). So whilst it would be good to get your changes in ASAP, there probably isn't a huge rush as it will be a few weeks (at least) before we are ready to publish the v18.0 release of fileserver.

conf/ftp
anandphulwani: @JedMeister Have pushed the changes.

JedMeister:
Hi @anandphulwani 

This is looking pretty good to me! Good work and thanks for your patience!

Last step is to get my colleague @OnGle to have a look and double check for anything that I might have overlooked. He's finished for the week, but hopefully will take care of that early next week.

JedMeister (approved):

anandphulwani:
Anytime, do tell me if you are thinking of some other improvements to be made.
As i said earlier i am really looking forward to create some fresh appliance, but maybe with the new release.

Again, cheers for the good work.

> Hi @anandphulwani
> 
> This is looking pretty good to me! Good work and thanks for your patience!
> 
> Last step is to get my colleague @OnGle to have a look and double check for anything that I might have overlooked. He's finished for the week, but hopefully will take care of that early next week.
> 
> (Note I have also asked @Christie018 to also have a look - but I'll probably only wait for @OnGle)

Christie018 (approved):
Looks like you guys have put a lot of work into this one. Looks good to me. Should be ready for testing then 👍🏾

anandphulwani:
@Christie018 Thanks for those kind words, just wanted this feature badly for myself. :)

OnGle (approved):
Great work, looks good to me :+1: Lets get this included for 18.x if we can!

(@JedMeister we'll need to add a fileserver common changelog with credit too)

JedMeister:
Thanks @Christie018 & @OnGle - I'll merge this ASAP.

Also, the PR needed to be retargetted against the `18.x` branch (rather than master). Unfortunately,. that has introduced a merge conflict. It should be easy to fix though, I'll take care of that via CLI shortly.

JedMeister:
I'm closing this PR. As noted in https://github.com/turnkeylinux/common/pull/274 I have rebased your work @anandphulwani so it will merge into the `18.x` branch cleanly.

Apologies that I didn't notice sooner. FWIW as of stable `18.x` release of TKLDev (almost any day now I hope...) the default branch will become `18.x`.

anandphulwani:
Thanks a tonne @JedMeister @Christie018 @OnGle

PR #248, opened 2023-06-02
Jed's 18.x dev

JedMeister:
First commit is a minor change to include new default initramfs compression tool (`zstd`).

The next 2 contain a fair bit of refactoring to `conf/bootstrap_apt` for bookworm (with intention to create minimal regressions for earlier releases now and in future).

- more detailed info re env vars (comments at top of script)
- require `RELEASE` (not `CODENAME` as it was previously)
- mechanism to pre-seed env vars via `/turnkey-buildenv` env file (each line should be `export ENVAR_KEY=value`)
- refactor previous release specific tweaks - to make it future transitions easier
- define `DEBIAN_PHP_V` programmatically (as it should have been before...)
- refactor @OnGle's work to use env var `NO_TURNKEY_APT_REPO` (via env var file `/turnkey-buildenv`)
- remove duplicate PHP pin for `*` in default Sury PHP config
- move consistent `PHP_EXTRA_PINS` packages to default Sury PHP config
- only add `PHP_EXTRA_PINS` if they don't already exist in default Sury PHP config

Note related to / required by: https://github.com/turnkeylinux/buildroot/pull/7

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -63,7 +63,7 @@ elif [[ $deb_ver -ge 11 ]]; then
     PROXY_PORT=3128
 fi
 
-if [[ "${APT_PROXY_OVERRIDE,,}" -eq "disable" ]]; then
+if [[ "${APT_PROXY_OVERRIDE,,}" == "disable" ]]; then
     PROXY_PORT=
 elif [[ -n $APT_PROXY_OVERRIDE ]]; then
     PROXY_PORT=$APT_PROXY_OVERRIDE
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -210,7 +210,7 @@ Pin-Priority: 550
 EOF
     # dynamically add some extra pins as specified in appliance Makefile
     for package_name in $PHP_EXTRA_PINS; do
-        if grep -q "Package: $package_name" /etc/apt/preferences.d/php-sury.pref; then
+        if grep -q "^Package: $package_name\$" /etc/apt/preferences.d/php-sury.pref; then
             continue
         fi
         cat >> /etc/apt/preferences.d/php-sury.pref <<EOF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -1,12 +1,28 @@
 #!/bin/bash -e
+
 # create apt sources
 # environment variables:
-#   - required: CODENAME
-#   - optional: NONFREE
-#   - optional: TKL_TESTING
-#   - optional: BACKPORTS
-#   - optional: PHP_VERSION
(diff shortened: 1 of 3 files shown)


Christie018 (approved):
Code changes to update common look good to me. My only question would be how does this refactoring affect future appliance updates that require nonstandard php versions?

JedMeister:
Thanks @Christie018 
> My only question would be how does this refactoring affect future appliance updates that require nonstandard php versions?

To be really explicit, the changes in this PR that directly related to alternate PHP install (all in `conf/bootstrap_apt` common conf script) are:
- define `DEBIAN_PHP_V` programmatically (as it should have been before...)
  - this shouldn't make any real difference other than it will be generated programmatically at build time (previously it was hardcoded). In theory that means that it should "just work" in future releases. In practice that may not be the case, but we'll cross that bridge if/when we get to it (when we do v19.x transition).

 - remove duplicate PHP pin for `*` in default Sury PHP config
   - this should make no functional change - there were just 2 duplicate `Package: *` entries...

 - move consistent `PHP_EXTRA_PINS` packages to default Sury PHP config
   - this should make no functional change. I just added the existing `PHP_EXTRA_PINS` packages (in individual apps' `Makefile`) to the default pinning file, generated at build time. FWIW we were adding the same 2 packages to all apps that included Sury PHP - so I just made them default part of the Sury pinning. The functionality of `PHP_EXTRA_PINS` remains essentially unchanged if we need it (although by default we shouldn't and existing ones should be removed for v18.x - see more below).

 - only add `PHP_EXTRA_PINS` if they don't already exist in default Sury PHP config
   - this change just eliminates the possibility of duplication when the `conf/bootstrap_apt` common conf script generates the package pinning file. To be clear, it won't stop us manually (accidentally) adding duplicate entries to `bootstrap_apt`, but it will ignore any `PHP_EXTRA_PINS` that already exist. So strictly speaking we don't need to remove the existing `PHP_EXTRA_PINS` (in the apps that already have it defined) because they will be duplicates (see point above). Obviously because it's redundant, we'll aim to remove the existing `PHP_EXTRA_PINS` (in individual apps) for v18.x.

OnGle:
@JedMeister in the future please link related PRs, in this case https://github.com/turnkeylinux/buildroot/pull/7 is required for the NO_TURNKEY_APT_REPO env-var name change.

OnGle (approved):
lgtm

JedMeister:
> @JedMeister in the future please link related PRs, in this case [turnkeylinux/buildroot#7](https://github.com/turnkeylinux/buildroot/pull/7) is required for the NO_TURNKEY_APT_REPO env-var name change.

Good call! Whilst it's a bit late now, I've made notes in the OP of both PRs cross-referencing one another.

PR #251, opened 2023-06-12
basic php wrapper script

JedMeister:
A variation on the request, but essentially resolves https://github.com/turnkeylinux/tracker/issues/1691.

I'm not sure if this is a good idea, or whether we should make it more Nextcloud specific?


OnGle (approved):
lgtm

PR #252, opened 2023-06-15
More 18.x dev bits & pieces

JedMeister:
A couple of random commits cleaning up common for v18.0

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -48,6 +48,9 @@ case $CODENAME in
         KEY_CODENAME=$CODENAME
         CONTRIB="contrib"
         NON_FREE="non-free"
+        ;;&
+    bookworm)
+        SEC_MIRROR="${SEC_MIRROR}debian-security"
         ;;
     # Note - only Ubuntu LTS
     focal|jammy)
@@ -140,6 +143,13 @@ cat > $SOURCES_LIST/sources.list <<EOF
 deb [signed-by=$key_dir/tkl-$KEY_CODENAME-main.gpg] http://archive.turnkeylinux.org/debian $KEY_CODENAME main
 
 deb $MIRROR_URL $CODENAME main
+EOF
+if [[ $deb_ver -ge 12 ]]; then
+    cat >> $SOURCES_LIST/sources.list <<EOF
+deb $MIRROR_URL $CODENAME non-free-firmware
+EOF
+fi
+cat >> $SOURCES_LIST/sources.list <<EOF
 deb $MIRROR_URL $CODENAME $CONTRIB
 #deb $MIRROR_URL $CODENAME $NON_FREE
 EOF
@@ -148,6 +158,13 @@ cat > $SOURCES_LIST/security.sources.list <<EOF
 deb [signed-by=$key_dir/tkl-$KEY_CODENAME-security.gpg] http://archive.turnkeylinux.org/debian $KEY_CODENAME-security main
 
 deb $SEC_MIRROR $sec_repo main
+EOF
+if [[ $deb_ver -ge 12 ]]; then
+    cat >> $SOURCES_LIST/security.sources.list <<EOF
+deb $SEC_MIRROR $sec_repo non-free-firmware
+EOF
+fi
+cat >> $SOURCES_LIST/security.sources.list <<EOF
 deb $SEC_MIRROR $sec_repo $CONTRIB
 #deb $SEC_MIRROR $sec_repo $NON_FREE
 EOF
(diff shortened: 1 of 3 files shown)


conf/bootstrap_apt
```diff
@@ -48,6 +48,9 @@ case $CODENAME in
         KEY_CODENAME=$CODENAME
         CONTRIB="contrib"
         NON_FREE="non-free"
+        ;;&
```
OnGle, line 51: From what I can tell you're trying to use fallthrough in this case statement, but afaik the correct syntax is `;&` rather than `;;&`

OnGle (commented):
Not sure if mentioned issue would cause bug or not. Otherwise looks fine.

conf/bootstrap_apt
JedMeister: Interesting. FWIW I checked the bash docs and TBH, I don't really understand the difference between `;&` & `;;&`!? Here's what [the manual](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Conditional-Constructs) says:

> [...] Using ‘;&’ in place of ‘;;’ causes execution to continue with the command-list associated with the next clause, if any. Using ‘;;&’ in place of ‘;;’ causes the shell to test the patterns in the next clause, if any, and execute any associated command-list on a successful match, continuing the case statement execution as if the pattern list had not matched.

My reading of that is that `;&` just keeps going through the rest of the case, whereas `;;&` does the same thing but described using more words!? What am I missing?

Out of interest, I made a quick test script and I honestly can't tell the difference between the behaviour?

JedMeister: To spell my tests out a bit more, here is `test-case.sh`:
```
#!/bin/bash -e

case $1 in
    a)
        echo 'a'
        ;&
    a|b)
        echo 'ab'
        ;&
    a|b|c)
        echo 'abc'
        ;&
    a|b|c|d)
        echo 'abcd'
        ;&
esac
```
And `test-case2.sh` is exactly the same, except uses `;;&` instead, i.e.:
```
$ diff -u test-case.sh test-case2.sh 
--- test-case.sh	2023-06-19 10:06:47.066076728 +1000
+++ test-case2.sh	2023-06-19 10:06:53.862232692 +1000
@@ -3,15 +3,15 @@
 case $1 in
     a)
         echo 'a'
-        ;&
+        ;;&
     a|b)
         echo 'ab'
-        ;&
+        ;;&
     a|b|c)
         echo 'abc'
-        ;&
+        ;;&
     a|b|c|d)
         echo 'abcd'
-        ;&
+        ;;&
 esac
```
And here is the output (testing 'a' - spoiler: the output is the same...):
```
$ ./test-case.sh a
a
ab
abc
abcd
```
```
$ ./test-case2.sh a
a
ab
abc
abcd
```
And if I use 'c' as the input (still get the same thing):
```
$ ./test-case.sh c
abc
abcd
```
```
$ ./test-case2.sh c
abc
abcd
```

JedMeister: Ok, so I've done a bit more looking at this and I have worked out the difference between `;&` & `;;&`. 

TL;DR:
 - `;&` - if matched, will "fall through" to the next option in order (without checking for a match). I.e. if all options use `;&`, then the first match will run, then **_all_** subsequent options will be run (i.e. regardless of whether they match or not).
 - `;;&` - will "resume" to allow subsequent matches. I.e. if all options use `;;&` then the first match will be run, then all subsequent options **_that match_** will be run (i.e. only if they match).

I've included a script which demonstrates all this pretty nicely below, but in context to your original question/feedback, either `;;&` or `;&` would work in the one you pointed out if `CODENAME=bookworm`, but would create a bug if `;&` was used and `CODENAME=buster|bullseye` (they would end up with bookworm's `SEC_MIRROR`). The other instance of `;;&` I used would also cause a bug if it were changed. Both `focal` and `jammy` would get `KEY_CODENAME="bullseye"`.

As such, I'm going to "resolve" this conversation and merge this PR.

Anyway, here's the script:
```
#!/bin/bash -e

for n in 4 14 24 34; do
    echo -n "$n = "
    case "$n" in
        3? )
            echo -n thirty-
            ;;&   #resume (to find ?4 later )
        "24" )
            echo -n twenty-
            ;&   #fallthru
        "4" | [13]4)
            echo -n four
            ;;&  # resume ( to find teen where needed )
        "14" )
            echo -n teen
    esac
    echo
done
```
Output (I added comments after the fact):
```
4 = four # only matches `"4"` (`four`)
14 = fourteen # matches `[13]4` (`four`) then "resumes" to match `"14"` (`teen`)
24 = twenty-four # matches `"24"` (`twenty-`) then "falls through" to `"4" | [13]4` (`four` - even though it doesn't match)
34 = thirty-four # matches `3?` (`thirty-`) then "resumes" to match `[13]4` (`four`)
```
If all the case options use `;;&` this is the result:
```
4 = four # as above
14 = fourteen # as above
24 = twenty- # matches `"24"` then "resumes" but there are no other matches
34 = thirty-four # as above
```
If all the case options use `;&`:
```
4 = fourteen # matches `"4"` (`four`) then "falls through" to `"14"` (even though it doesn't match)
14 = fourteen # as per original example (except "falls through" rather than "resuming")
24 = twenty-fourteen # matches `"24"` (`twenty-`) then "falls through" to `"4" | [13]4` (`four` - although it doesn't match) and again falls through to `"14"` (`teen` - again doesn't match)
34 = thirty-twenty-fourteen # matches `3?` (`thirty`), then "falls through" to `"24"` (`twenty-` - even though it doesn't match), then "falls through" to `"4" | [13]4` (`four` - matches `[13]4` - but it's irrelevant) and finally "falls through" to `"14"` (`teen` - which again doesn't match)
```

OnGle: Wow couldn't even find `;;&` on the docs, must've been blind. That said you're absolutely right and `;;&` is correct here.

PR #254, opened 2023-06-20
remove stunnel & shellinabox

OnGle:


JedMeister:
Awesome thanks man. However, I did a `grep -R` for `stunnel` and found a few more (plus `stunnel4` itself in the base plan):
```
root@tkldev6 .../fab/common# grep -R --exclude-dir=.git --exclude-dir=.mypy_cache stunnel .
./conf/turnkey.d/shellinabox:# (for shellinabox stunnel conf, see common shellinabox overlay)
./conf/turnkey.d/shellinabox:# via a stunnel.conf file - see shellinabox and/or stunnel common overlays for
./conf/turnkey.d/shellinabox:systemctl enable stunnel4@shellinabox.service || true
./plans/turnkey/base:stunnel4                /* provide ssl to shellinabox */
```
Also, while double checking, I also noticed a Webmin related tweak (in `zz-ssl-ciphers`) that should also be removed - see here: https://github.com/turnkeylinux/common/blob/2438bdb8f17d4464395a8e145c1f1b9c77037d52/conf/turnkey.d/zz-ssl-ciphers#L66-L70

I've opened a new branch with my suggested changes (on top of your commit): https://github.com/turnkeylinux/common/pull/255

PR #262, opened 2023-07-24
More 18.x fixes/improvements

JedMeister:
`fail2ban`
- provide default `jail.local` config file (rather than changing Debian supplied `jail.conf`)
- install `python3-systemd` package - finalises https://github.com/turnkeylinux/tracker/issues/1843, https://github.com/turnkeylinux/tracker/issues/1844 & https://github.com/turnkeylinux/tracker/issues/1850
- block known users using wrong keys
- weekly vacuum DB to keep filesize down
- explicitly set `allowipv6 = auto` (quiet logs - otherwise is warning is logged)

`postfix`
- ensure `postfix` starts

`sshd`
- replace common `sshd` conf script with conf file overlay (`/etc/ssh/sshd_config.d/turnkey.conf`) - improved user experience when upgrading ssh (not changing default config file will avoid asking user if/when default config file updated)

### DIFF
--- a/conf/turnkey.d/fail2ban-fixes
+++ b/conf/turnkey.d/fail2ban-fixes
@@ -1,8 +1,5 @@
 #!/bin/bash -e
 
-# make fail2ban default to reading journal - see #1844 (Debian #1037437)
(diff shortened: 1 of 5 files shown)


OnGle (approved):
nice, good stuff

PR #265, opened 2023-07-29
Disabling protocols for security hardening

Christie018:
Disabled tipc, dccp, sctp, and rds in common/turnkey.d/overlays to reduce the attack surface of all TKL appliances.

### DIFF
--- a/overlays/turnkey.d/protocol-hardening/etc/modprobe.d/rds.conf
+++ b/overlays/turnkey.d/protocol-hardening/etc/modprobe.d/rds.conf
@@ -1,0 +1,1 @@
+install rds bin/true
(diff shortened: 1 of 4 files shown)


overlays/turnkey.d/protocol-hardening/etc/modprobe.d/rds.conf
```diff
@@ -0,0 +1 @@
+install rds bin/true
```
OnGle, line 1: *probably* doesn't matter if you've checked it, but this is a relative path, might be best to add the leading slash

OnGle (changes requested):

overlays/turnkey.d/protocol-hardening/etc/modprobe.d/rds.conf
JedMeister: Good catch @OnGle

@Christie018 - could you please make that change and double check that it works as expected. Thanks.

PR #268, opened 2023-08-01
Securing appliances by disabling protocols

Christie018:
Disabled tipc, dccp, sctp, and rds in common/turnkey.d/overlays to reduce the attack surface of all TKL appliances.


OnGle (approved):
whooops sorry I missed this, looks good to me

PR #266, opened 2023-07-31
18.x common tweaks

JedMeister:
[update] Note `turnkey-composer` updates have been spilt off into their own PR - see #272 

Also, whilst the PR notes changes to `turnkey-php` that no longer appears to be the case...

----

<s>Make turnkey-composer better:
- add `-h|--help` arg
- make fallback defaults more obvious and more easily configured by end user
- support setting `APPROOT` (default cwd to run composer)

The intention is that when used in a specific appliance, the `APPR_FALLBACK` is updated to be the application root dir (and other defaults as relevant - mostly won't need to be touched).</s>

<s>And added some other</s> random tweaks:
- minor tweak to postfix config (just avoids warning message) - https://github.com/turnkeylinux/common/pull/266/commits/c346c92c18384b0f38ded6693ebb8588646cb8b3
- <s>improve `turnkey-php` - https://github.com/turnkeylinux/common/pull/266/commits/1de09576981c5b092ca42083a63914612460ca80</s>
- Add Core v18.0 changelog to common (`changes/turnkey.changelog`) - https://github.com/turnkeylinux/common/pull/266/commits/31314f23735e3eb1e7e818f185eefcff4bc8110b
- improve `service` wrapper script (include webmin & kill cleanly) - https://github.com/turnkeylinux/common/pull/266/commits/4db2633971ef9490f866072b88cb0c80d3d647d5

PR #272, opened 2023-08-30
Just the composer updates

JedMeister:
Make turnkey-composer better:
- add `-h|--help` arg
- make fallback defaults more obvious and more easily configured by end user
- support setting `APPROOT` (default cwd to run composer)

The intention is that when used in a specific appliance, the `APPR_FALLBACK` is updated to be the application root dir (and other defaults as relevant - mostly won't need to be touched).

Note that this PR was previously part of #266 but I split it out so I could merge that PR. @OnGle will look at this one a bit closer...


OnGle (approved):
lgtm!

PR #277, opened 2023-09-24
rails v18.0 (common)

qq7:


JedMeister:
Looks good mate, thanks! :)

JedMeister:
Also, this probably isn't the "right" place to mention this and you've likely already noticed but just in case, you can now set the desired Ruby version in the appliance Makefile - like this:
```
RBY_VER=3.0
```
(It passes the version string to `rbenv` during setup)

Also Node version can be set in a similar way, e.g. `NODE_VER=lts` (version string passed to `n`).

PR #285, opened 2023-12-20
update conf/web2py

a3s7p:
1. remove obsolete python2 compat
2. improve release query logic
3. perform a recursive shallow clone

### DIFF
--- a/conf/web2py
+++ b/conf/web2py
@@ -9,16 +9,11 @@ SRC=/usr/local/src
 W2PPASS=turnkey
 W2PROOT=/var/www/web2py
 WSGI=$W2PROOT/wsgihandler.py
-RELEASES='https://api.github.com/repos/web2py/web2py/releases/latest'
 
 # clone web2py
 cd /var/www
 
-TAG="$(
-    curl "$RELEASES" -s \
-    | grep --color=never '"tag_name":' \
-    | grep -o '[0-9][^"]\+'
-)"
+TAG="$( gh_releases web2py/web2py | tail -1 )"
 
 git clone \
     --depth 1 \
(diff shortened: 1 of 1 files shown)


conf/web2py
```diff
@@ -9,38 +9,29 @@ SRC=/usr/local/src
...
 
-# Install GitPython in virtualenv to workaround turnkey-pylib
-# git.py conflict
-virtualenv -p python2 $W2PROOT/venv2
-$W2PROOT/venv2/bin/pip install GitPython
+TAG="$(
+    curl "$RELEASES" -s \
+    | grep --color=never '"tag_name":' \
+    | grep -o '[0-9][^"]\+'
+)"
```
JedMeister, line 21: This is pulling in "release" `2.19.1` : https://github.com/web2py/web2py/releases/tag/2.19.1 (from Mar 22, 2020).

But latest is `v2.26.1`: https://github.com/web2py/web2py/releases/tag/v2.26.1 (from Oct 30, 2023)

Was using that much older version intentional?

FWIW it looks like they used to do "releases" but `2.19.1` was the last one. It looks like they've switched to just using tags (and have added a leading 'v').

FWIW, this gives the correct tag:

```
gh_releases web2py/web2py | sort -V | tail -1
```

I.e.:
```
root@tkldev8 ~# TAG=$(gh_releases web2py/web2py | sort -V | tail -1)
[WARNING] GITHUB_USER not set! GITHUB_USER_TOKEN not set! Authentication won't be used.
Fetching releases from github for 'web2py/web2py'... Done!
root@tkldev8 ~# echo $TAG
v2.26.1
```

Obviously you could tweak your code to achieve the same ends, but `gh_releases` (in common) takes care of it too.

JedMeister:
LGTM with exception of my comment. Unless there is some reason not to pull the latest version, that will need tweaking.

a3s7p:
Of course you're right, somehow I missed that.

Fixed & tested to work, also with clicksnap.

It looks like the `sort -V` in the pipeline is not needed since it's called in `gh_releases` already.

PR #286, opened 2024-01-11
Rails common conf - (properly) fallback to latest stable if not set

JedMeister:
TBH, I'm not 100% sure if this is still required or not, but I just noticed it uncommitted in a local TKLDev, so figured I'd commit it and add a PR.


OnGle (approved):
lgtm

PR #289, opened 2024-01-23
insert tkl-upgrade-php.sh

marcos-mendez:
I added the file inside the php folder as this will be compiled by any PHP application (either nginx, apache or php-fdm)

Please correct me if i was wrong and in case i need to change anything else

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "overlays/php/usr/local/bin/tkl-upgrade-php"]
+	path = overlays/php/usr/local/bin/tkl-upgrade-php
+	url = https://github.com/popsolutions/tkl-upgrade-php/


marcos-mendez:
I am not really used to compiling the tkl common repo, i assume it is compiled everytime a tkl-core is made or something like that. I couldn't compile it because of ignorance on how i should be performing this. Anyways if the PR is wrong feel free to tell me where i should put it. Also notice that i tought that php folder was the best idea but it can be the worst one, i don't know.

Also, i added it as a submodule. I don't know if this is a best practice in the TKL community altought in the Odoo community i am more used to is how the do it. 

If the refrs git has to be changed please let me know

marcos-mendez:
Ok, so i just push the script inside the php folder and remove the
submodule is that right?

Em ter., 23 de jan. de 2024 às 00:49, Jeremy Davis ***@***.***>
escreveu:

JedMeister:
Yeah mate, that'd be awesome thanks. Although as I said, if you don't get a chance, then I'm happy to do it.

Although it looks unlikely that I'll get to it this week. If it doesn't get in this week, then it's unlikely to make it into the next batch (of ~10) updated apps I'm hoping to get completed.

Regardless, it'd still be of huge value as I can still direct users of current releases to manually download.

marcos-mendez:
Jed, do you think i can work for TurnKey Linux? Lol if you techme how i
could start helping you. I really love you and your project. I don't know
how much important you are here in Brazil.

About the script i'll put it right away. I make a new Pull Request

Em ter., 23 de jan. de 2024 às 19:26, Jeremy Davis ***@***.***>
escreveu:

PR #297, opened 2024-02-24
systemd-chroot: fix buggy service wrapper script

a3s7p:
The `service` wrapper has been buggy since 1d459feb.

1. The order of arguments for `stop_start` is wrong for: gitlab-runsvdir, ghost, mattermost, redis.
    - It's `$1 $2` (`<service name> <action>`) instead of `$2 $1` (`<action> <service name>`) which the function expects.
    - The correct order is used for webmin which was added by 44a90b93.
2. The script fails silently which is probably the reason why the issue was not detected earlier.
    - By calling `stop_start` in the background with `&`, no exit code is returned unless one `wait`s for it later.
    - There is no `wait` later.
    - Therefore the script exits 0 even though `stop_start` failed immediately.

This commit fixes the issue which prevents e.g. the Gitlab appliance from building.

Misc fixes:
- apply shellcheck lints
- use consistent glob for PHP version

### DIFF
--- a/overlays/turnkey.d/systemd-chroot/usr/local/bin/service
+++ b/overlays/turnkey.d/systemd-chroot/usr/local/bin/service
@@ -5,26 +5,35 @@
 # Should deal with common build time calls to "service" with common actions
 # If the service isn't recognised, then will fall back to /usr/bin/service
 
-
 stop_start() {
-    _action=$1 # stop|start|restart
+    _action="$1" # stop|start|restart
     shift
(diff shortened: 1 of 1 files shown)


JedMeister:
Oops! My bad.. Gross oversight and insufficient testing. thanks for fixing @a3s7p.

FWIW I've now got shellcheck running in vim (albeit only locally - not on my tkldev - yet).

Anyway, I haven't tested this, but happy to take your word for it (what you say looks right on face value). As such now merging...

PR #301, opened 2024-05-01
Ruby/Rails common stuff - bugfixes etc

JedMeister:
Most this is bugfixes and shellcheck implementations, although some minor refactoring.

TBH, it seems to me that the common parts of these scripts should be split out into a `rails-common` - with just the specific mysql/mariadb and postgres bits in the other scripts. And probably rename the (mysql/mariadb specific) from `rails` to `rails-mariadb`. But that's for another day...


OnGle (approved):
lgtm

PR #303, opened 2024-05-14
Clean up old MariaDB bugfixes - no longer required

JedMeister:
General improvements; somewhat related to https://github.com/turnkeylinux-apps/redmine/pull/29 - and also perhaps related to https://github.com/turnkeylinux/tracker/issues/1947 ?

The removal of `/var/lib/mysql/ib_logfile*` files was a bugfix for a weird issue **13 years ago**! I have just done a build without removing these files and it appears to be fine.

The `innodb-barracuda.cnf` file was to work around issues with utfmb4 on an older version of MariaDB (that didn't default to supporting it). The settings in it are now either deprecated, or the defaults.


OnGle (approved):
lgtm!

PR #307, opened 2024-06-12
Improve/fix turnkey-artisan:

JedMeister:
Closes https://github.com/turnkeylinux/tracker/issues/1959

Fixes failing usage of `turnkeyartisan` usage in Usahidi:
https://github.com/turnkeylinux-apps/ushahidi/blob/master/conf.d/main#L56

- cd to WEBROOT first - Ushahidi failing build time fix
- use absolute path to php - so it will run properly if used in cron job
- apply shellcheck suggestion - even though probably irrelevant


OnGle (approved):
Looks good to me

PR #309, opened 2024-08-30
Fix php-fpm in dummy service script used during build

JedMeister:
Not sure how this happened, but the `redis` function in our build time dummy service script was trying to start php-fpm!?

This PR fixes that. Although I note that with this fixed, there isn't any handling of redis now (even though it was wrong). So perhaps redis needs an entry? Having said that, I'm assuming that it doesn't as I'm sure that it would have caused issues before now otherwise.


OnGle (approved):
Huh... weird, well looks good to me

PR #313, opened 2025-01-06
Update tkl-upgrade-php to php 8.4

marcos-mendez:
The script is working from low to high so im keeping this architecture for nextcloud (as it is already working)

### DIFF
--- a/overlays/php/usr/local/bin/tkl-upgrade-php
+++ b/overlays/php/usr/local/bin/tkl-upgrade-php
@@ -9,7 +9,8 @@ choose_php_version() {
     echo "4) PHP 8.1"
     echo "5) PHP 8.2"
     echo "6) PHP 8.3"
-    read -r -p "Enter choice [1-6]: " php_choice
+    echo "7) PHP 8.4"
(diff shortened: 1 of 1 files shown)


marcos-mendez:
This should be review by this time in the current PHP. As im not a PHP programmer only a monkey-patcher this is the suggestion i have.

Here is the refferance

https://www.php.net/supported-versions.php

JedMeister:
Thanks @marcos-mendez. That's great!

JedMeister:
I agree that keeping it up is an important thing to do. And having some facility to automate that would be ideal. I think that it would also be a good idea to move this script from common and package it instead would make that better. I've opened a feature request so that idea doesn't get forgotten: https://github.com/turnkeylinux/tracker/issues/2023

JedMeister:
Actually, I think it would be better if the script would dynamically update itself. That would reduce manual update and not require this script to be packaged - although it's probably a good idea to package it anyway.

I've opened a separate issue for that: https://github.com/turnkeylinux/tracker/issues/2024

PR #314, opened 2025-02-18
18.x arm64

a3s7p:
- **plans: update, move rpi-specific changes to bt**
- **overlays: move raspi-specific to bt patch**
- **conf: move rpi-specific to bt patch**


a3s7p:
Related to https://github.com/turnkeylinux/fab/pull/30 https://github.com/turnkeylinux/buildtasks/pull/92

PR #315, opened 2025-02-27
Major refactor and tidy of `tkl-composer-squash-vendor`

JedMeister:
Fix and lint `tkl-composer-squash-vendor` script.

Closes https://github.com/turnkeylinux/tracker/issues/2036

PR #318, opened 2025-03-19
fix not finding vendor

OnGle:
`tkl-composer-squash-vendor` currently mishandles the vendor list, returning all directories ending with "vendor" in a newline separated list, causing it to fail the following `-d` check. This simplifies the code and fixes this issue.

noted here https://github.com/turnkeylinux/common/pull/315


JedMeister:
Awesome, thank you! :)

PR #320, opened 2025-06-03
upgrade tkl-upgrade-php script

OnGle:
- show version support status
- show which version is provided by debian
- install debian version when switching to version provided by debian


JedMeister:
Nice! :heart:

PR #323, opened 2025-07-09
PHP sury tweak

JedMeister:
- add initial support for trixie (may need more)
- remove `buster` as an explicitly supported codename - some redundant code remains
- apply shellcheck lints
- fallback to system if `RELEASE` not set
- rename sury.org gpg key file to make it consistent with upstream (see https://packages.sury.org/php/README.txt)


OnGle (approved):
lgtm

PR #324, opened 2025-08-07
fix our WebDav-CGI logout not working as intended

JedMeister:
As explained in https://github.com/turnkeylinux/tracker/issues/2058#issuecomment-3162684474 our custom logout button in WebDav-CGI wasn't working properly.

Using a blank username resolves that.

However, IMO the root cause of the issue is that logging in with any username other than a valid samba user results in login as the guest samba user...


OnGle (approved):
lgtm!

PR #325, opened 2025-08-12
Fix buggy sed (wasn't picking up message) - just dump the whole json …

JedMeister:
Just hit the GH API limit (unauthenticated) and was struggling to work out what was going on!

Turns out the sed string was buggy... I was going to fix it, then thought why? It's a dev tool so best to just dump the json and be done with it...


OnGle (approved):
Nice, perfectly reasonable solution

PR #328, opened 2025-10-08
Trixie apt setup - use deb822 format

JedMeister:
At the time of writing this isn't quite right and is not working - but hopefully by the time you look at this @ongle it should be good to go...


OnGle (approved):
lgtm!

PR #329, opened 2025-10-31
Config related v19.x/Trixie updates - mostly webserver and security related

JedMeister:
- some minor linting related updates (mostly shellcheck)
- webservers (apache, nginx & lighty)
    - cipher updates
    - disable access to all dot files by default
    - other security related conf hardening
    - other general conf refactoring (mostly apache)
- tomcat
    - cipher updates
    - other conf updates, mostly aesthetic
    - update java version
    - add support for tomcat 11 - Trixie has both 10 & 11 - currently support for both is in a single overlay, but perhaps they should be split up?
- postfix
    - cipher and other config updates
    - manually start postfix at build time - init.d script now fails in chroot (perhaps should be moved to our dummy systemctl script?)
- sysctl config
    - move from conf script that appends to default conf to separate purpose specific overlays in /etc/sysctl.d/ dir
- plan adjustments
    - updates for changed Trixie packages
    - temporarily comment di-live & tklbam


OnGle (approved):
lgtm!

PR #331, opened 2025-12-03
Explictly set supported arch in sources files

JedMeister:
Strictly speaking this isn't necessary, but is a nice touch. It means that systems with multiarch enabled (e.g. i386 on an amd64 system - like my laptop) will not bitch about the TKL repo not supporting the alternate arch (i386 in my case).


OnGle (approved):
lgtm!

PR #333, opened 2026-01-22
Misc fixes/cleanup

JedMeister:
- Cleaner method to kill gpg-agent
- Fix malformed sources file
- Handle proxy related vars


OnGle (approved):
lgtm

PR #334, opened 2026-01-23
Trixie Tomcat updates - add config for Tomcat 11

JedMeister:
Trixie has both Tomcat 10 & 11 - this is new config for 11, leaving the existing 10 conf in place. Please note that it has not been significantly tested.

TBH I wasn't sure whether we should make v10 & v11 separate, but I think that it's probably best to just bundle them together for now and just release tomcat11. If there is demand, perhaps we could build in support for changing to the other (i.e. 11 -> 10)?

Open to further thoughts..

### DIFF
--- a/overlays/tomcat-apache/etc/tomcat11/tomcat11/mod_jk.conf
+++ b/overlays/tomcat-apache/etc/tomcat11/tomcat11/mod_jk.conf
@@ -1,0 +1,7 @@
+RedirectMatch ^/manager[/]$ /manager/html
+JkMount /manager         ajp13_worker
+JkMount /manager/*       ajp13_worker
+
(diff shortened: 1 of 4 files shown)


OnGle (approved):
Seems a bit odd to me to keep the old tomcat config but not have a method to swap back. But if it doesn't cause any issues then I don't see any actual issues with it. lgtm!

JedMeister:
> Seems a bit odd to me to keep the old tomcat config but not have a method to swap back. But if it doesn't cause any issues then I don't see any actual issues with it. lgtm!

In essence I agree, but I figure it gives us some future flexibility - i.e. things like building both version variants as separate appliances, provide a "downgrade tomat" script or even if we have a tomcat appliance that requires a different tomcat version we have reference conf built-in.

PR #336, opened 2026-02-26
Mysqltuner fix

JedMeister:
This PR primarily addresses https://github.com/turnkeylinux/tracker/issues/2084 (`mysql` conf script bug) - via commit https://github.com/turnkeylinux/common/commit/77763b4bdff8de86001db6be8879b529a36a7a97.

It also includes conversion to bash and shellcheck fixes - in https://github.com/turnkeylinux/common/commit/87ea9642a313021d426c21fa17d2c9e20d266780.

However as noted in the above bug; another curl related bug also applies to the download failure - see https://github.com/turnkeylinux/tracker/issues/2083. As noted in that issue (as a "feature request" note towards the bottom) it might be best to create a standalone download script that wraps around curl. So rather than using my `dl()` function updates here, it would probably be better to just write that script - then replace all individual `dl()` functions (e.g. in appliances) everywhere.

### DIFF
--- a/conf/mysql
+++ b/conf/mysql
@@ -1,9 +1,13 @@
-#!/bin/sh -e
+#!/bin/bash -e
 
 # download mysqltuner
 dl() {
-    [ "$FAB_HTTP_PROXY" ] && PROXY="--proxy $FAB_HTTP_PROXY"
-    cd $2; curl -L -f -O $PROXY $1; cd -
+    if [[ "$FAB_HTTP_PROXY" ]]; then
+       PROXY=(--proxy "$FAB_HTTP_PROXY")
+    fi
+    cd "$2"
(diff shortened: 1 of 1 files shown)


JedMeister:
Also @ongle - if you like my idea (not using the `dl()` function updates in the second commit) I might just merge the first commit. Thoughts?

OnGle (approved):
Lgtm!

OnGle:
> Also @OnGle - if you like my idea (not using the `dl()` function updates in the second commit) I might just merge the first commit. Thoughts?

Doesn't bother me eitherway. The work's already been done and the seperate DL script will take a while to test and generalize across all usages in our confscripts. May as well include the second commit and get it working for everyone so we can get these early lamp images out asap.

PR #338, opened 2026-03-24
Add dual-stack IPv4/IPv6 network support for TKL v19

marcos-mendez:
### 1. Replace udhcpc with dhcpcd for dual-stack DHCPv4/DHCPv6
- Switch from udhcpc (IPv4-only) to dhcpcd (dual-stack v4+v6)
- Add inet6 dhcp stanzas to interfaces overlay
- Remove udhcpc-fix overlay
- Mask dhcpcd.service daemon to prevent conflict with ifupdown

### 2. Enable apache-ssl conf script for all Apache appliances
- apache-ssl existed in common/conf/ but was never invoked by any .mk
- Add apache-ssl to apache.mk COMMON_CONF

### Testing
- Built and tested Moodle appliance as LXC on Proxmox
- Verified dual-stack networking (DHCPv4 + DHCPv6/SLAAC)
- Verified Apache starts correctly with SSL certificate

PR #337, opened 2026-03-12
Apache & Mysql Fixes

OnGle:

PR #339, opened 2026-04-02
TKL v19: dual-stack networking, Redis/Ruby verticals, and misc fixes

marcos-mendez:
## Summary
- Dual-stack IPv4/IPv6 network support (dhcpcd replacing udhcpc)
- Fix DHCPv6: remove redundant inet6 dhcp from interfaces
- Add Redis vertical: plan and conf for shared Redis infrastructure
- Add Ruby vertical: extract Ruby build deps from Rails into standalone plan
- Minor NGINX overlay SSL snippet fix
- Apache SSL conf script fix, mawk/locales base plan fixes, fail2ban guard, PostgreSQL plan

Replaces #338 (was opened from the wrong fork `popsolutions/tkl-common`; now from `popsolutions/common`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JedMeister:
After a bit of testing, I'm going to close this PR.

I've  opened #343 as a partial replacement that includes your idempotency improvements - but **NOT** the DHCPv6 stuff. (It also doesn't include [e34762f](https://github.com/turnkeylinux/common/pull/339/commits/e34762f2421d58dc13eb7e48a5c6291d61151a78) - updates to `mk/turnkey/apache.mk` - because that was also done in #337).

I'm on board with improving DHCPv6 support but testing your changes I found IPv4 DHCP unreliable and setting static IPv4 IPs via Confconsole nonfunctional. After a bit of research I made a few discoveries that I think may explain my experience - primarily that `dhcpcd` is not compatible with `ifupdown` & use of the `/etc/network/interfaces` file. Rather than detail all that here though, I'm going to open a new issue on the tracker so we can discuss it further there.

Because implementation of IPv6 requires a bit more consideration it's likely we may not be able to implement full IPv6 DHCP support in v19.0. Because v19.0 is so overdue already, I'd rather release something that is sub-optimal - but with the same functionality as v18.x - than something that includes new features which potentially cause new issues.

JedMeister:
Oops, hit the wrong button...

PR #340, opened 2026-04-09
Plan updates

JedMeister:
The primary point of this PR is to update the base plan to include `tkl-installer` and completely remove `di-live`. As you can see I also did some other tidying up. TKLBAM is now also enabled - however it's not well tested. Plus it adds a fair bit extra to the ISO - mostly because of it's dependency on `python3-botocore` - which includes support for every AWS service known to humankind... I have a plan to slim that down, but I'll get to that when I get a chance.

The (conf & overlay) readmes also included here aren't really relevant - or that useful - until we implement/finalize `arm64` support. But I had already done them so figured I might as well include them...

### DIFF
--- a/conf/turnkey.d/README.md
+++ b/conf/turnkey.d/README.md
@@ -1,0 +1,27 @@
+Common TurnKey configuration scripts
+====================================
+
+These configuration scripts will be applied to all TurnKey appliances.
+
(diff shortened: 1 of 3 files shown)


OnGle (approved):
Looks great, can almost taste the 19.x RC builds!

PR #341, opened 2026-04-09
revert recent changes to conf/bootstrap_apt - causes issues with recent fab update

JedMeister:
I haven't bothered looking into it, but with the changes to `conf/bootstrap_apt` in combination with the recent fab changes, `NO_PROXY=true` is broken. I've just tested this and reverting to the previous iteration makes it work again (with latest fab and current `19.x-dev` common).

Reverting the `PROXY_PORT=8124` line wasn't really necessary but as it only applies to Debian 9 and older it's irrelevant either way IMO...

### DIFF
--- a/conf/bootstrap_apt
+++ b/conf/bootstrap_apt
@@ -26,6 +26,9 @@
 #       - if not the same as guest, apply relevant transition changes
 #   - NO_TURNKEY_APT_REPO <optional>:
 #       - disable TurnKey apt repos - useful during early transition
+#   - NO_PROXY <optional>:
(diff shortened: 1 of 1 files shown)


OnGle (approved):
Looks good, I'm gonna approve this for now but tomorrow I'll see if I can't dig up my earlier apt changes that didn't land and see if I can get any insight on this.

JedMeister:
FYI:
```
root@chill-time:/turnkey/fab/products/tkldev# apt list fab
fab/trixie-testing,trixie-testing,trixie-testing,now 1.1rc5 all [installed]
```

PR #342, opened 2026-04-09
Extend git aliases and enable autocomplete

JedMeister:
This is not at all important but it's what I use on my local PC so required zero effort. :grin: 

It extends on the existing `~/.bashrc.d/git` file; adds some additional aliases and provides autocomplete.

It also now passes shellcheck and prefers long options over short ones, which IMO makes it easier to see what stuff does at a glance - particularly for options that you may not be familiar with.

### DIFF
--- a/overlays/turnkey.d/bashrc/etc/skel/.bashrc.d/git
+++ b/overlays/turnkey.d/bashrc/etc/skel/.bashrc.d/git
@@ -1,23 +1,51 @@
-# uncomment if not setting 'git config'
+# uncomment these or use 'git config'
 #export GIT_AUTHOR_NAME="Your Name"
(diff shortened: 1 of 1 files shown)


OnGle (commented):
Great addition, but I'm a bit mixed on adding all of these in common for all turnkey users by default. Especially since not all of our users will be using git, and these alias' aren't publicly documented anywhere.

JedMeister:
Fair call. I think at least some of the improvement/changes (e.g. shellcheck updates) should be included, but it's low priority and I'm not attached so I'll close this and open a new one against TKLDev explicitly.

Replaced by TKLDev specific overlay: https://github.com/turnkeylinux-apps/tkldev/pull/47

PR #347, opened 2026-04-15
fix: use pg_ctl directly in chroot for reliable PostgreSQL start/stop

marcos-mendez:
## Summary

`/etc/init.d/postgresql stop` and `pg_ctlcluster stop` both fail with "pid file is invalid" inside the FAB chroot because `/proc` is shared with the host. The PID file written by PostgreSQL contains the host-namespace PID, which the stop wrappers cannot validate against the chroot's `/proc`.

## Impact

The `conf/pgsql` script fails during appliance builds with:
```
Stopping PostgreSQL 17 database server: mainError: pid file is invalid,
please manually kill the stale server process. ... failed!
```

This blocks all appliance builds that use PostgreSQL.

## Fix

Replace `/etc/init.d/postgresql start/stop` with direct `pg_ctl` binary calls via `su postgres -c "/usr/lib/postgresql/VER/bin/pg_ctl ..."`. The `pg_ctl` binary manages the process directly without PID file validation wrappers.

## Testing

Full NetBox appliance build completes successfully with PostgreSQL 17 — cluster creation, password configuration, user creation, and clean shutdown all work correctly inside the FAB chroot.


JedMeister:
Hey mate. Good catch re the postgres issue!

I do recall @OnGle mentioning that there were issues with the postgres common code earlier, and that he had fixed it/them. I suspect that it was the same issue you note here? But I can't see any related recent commits in the history? Perhaps he forgot to push his changes?

I'll get @OnGle to double check if he has a local fix that he forgot to open a PR for as a priority. If he has already done the work, we'll pull that in ASAP - should be today.

---

FYI re starting/stopping services, instead of updating the common postgres conf script to start/stop `pg_ctl` directly, I'd rather the conf script be updated to use our fake build time [`systemctl`](https://github.com/turnkeylinux/common/blob/18.x/overlays/turnkey.d/systemd-chroot/usr/local/bin/systemctl)/[`service`](https://github.com/turnkeylinux/common/blob/18.x/overlays/turnkey.d/systemd-chroot/usr/local/bin/service) scripts instead. They're a bit hacky, but we include them because some software explicitly calls `systemctl`/`service` during install and it's nice to have a common place for all those service start/stop calls to be managed in a consistent way.

E.g in the case of the issue discussed here; add `postgrsql) ...` to the [`case` statement in the fake `service` command](https://github.com/turnkeylinux/common/blob/18.x/overlays/turnkey.d/systemd-chroot/usr/local/bin/service#L48).

Then change lines like this:
```
/etc/init.d/postgresql start
```
to
```
systemctl start posgresql
```
(you could use our fake `service` script directly in the conf script instead, although my preference is to use the fake `systemctl` script where possible - which calls the fake `service`).

Hopefully that all makes sense?! If not please let me know and I'll try again.

---

Regardless of that I possibly would have just merged a fix for the issue as you described, but there are a few issues with this PR:

It was against `master` instead of `19.x-dev`. I've taken care of that, so no problem. But then I also noticed that this PR also includes some commits irrelevant to this issue that haven't been merged into `19.x-dev` yet.

And I can't see a commit that applies the fix for the issue you mention? Don't worry about that for now, I'll check @OnGle

---

And as a more general dev request, could you please try to do future PRs are on top of a "clean" current `19.x-dev`? If your local dev branch diverges a little bit and/or is a few commits behind the current `19.x-dev` is not a big deal because rebasing/cherry-picking a few commits on my end and opening a new PR is usually pretty easy. But the more the basis of a PR has diverged from `19.x-dev` common, the more work it is for me to work out which specific commits to cherry pick for the specific changes from your PR and then fix any merge conflicts.

If you want to retain your own local dev branch as you have changes that we haven't merged yet, that's completely reasonable. But it would reduce my workload if you could fork a new branch from current `19.x-dev` - and save me having to do it! :grin:. E.g.:

```
# assuming 'turnkey' is the remote name for 'github.com/turnkeylinux/common' & you start on a local dev branch

# show the last few commits so you know the IDs of the commits to include in your PR; e.g. list last 10 commits
git log --oneline -10

# checkout a new "bugfix-postgres" branch based off current turnkey 19.x-dev
git checkout -b bugfix-postgres -t turnkey/19.x-dev

# cherry pick the specific commit/s you want in your new PR - e.g. b4605ba
git cherry-pick b4605ba

# sometimes you may need to resolve a conflict
vim some/file # fix conflict
git add some/file
git cherry-pick ---continue
```

When I have some time, I should document and clarify my dev git workflow better, but in the meantime if you have specific questions please ask.

marcos-mendez:
Hi @JedMeister, thanks for the detailed feedback — it makes total sense!

I've opened #348 as a clean replacement for this PR:
- **Clean branch** based on current `19.x-dev` — only the PostgreSQL fix, no unrelated commits
- **Uses the fake `systemctl`/`service` infrastructure** as you suggested — added a `postgresql)` case to the fake `service` script that calls `pg_ctl` directly via `su postgres`, and updated `conf/pgsql` to use `systemctl start/stop/restart postgresql`

This way PostgreSQL service management during builds follows the same pattern as apache2, postfix, mariadb, etc.

I'll also follow the cherry-pick workflow you described for future PRs — fork a clean branch from `19.x-dev` and cherry-pick only the relevant commits. Thanks for taking the time to explain the process!

Feel free to close this PR in favor of #348.

marcos-mendez:
Closing in favor of #348 — a clean PR based on 19.x-dev with only the PostgreSQL fix, using the fake systemctl/service infrastructure as suggested by @JedMeister.

JedMeister:
Love your work - thanks mate!

PR #348, opened 2026-04-16
fix: route PostgreSQL start/stop through fake systemctl/service in chroot

marcos-mendez:
## Summary

- Add `postgresql)` case to the build-time fake `service` wrapper (`overlays/turnkey.d/systemd-chroot/usr/local/bin/service`) that calls `pg_ctl` directly, bypassing the PID-file validation that breaks inside the FAB chroot
- Update `conf/pgsql` to use `systemctl start/stop/restart postgresql` instead of `/etc/init.d/postgresql` calls

## Problem

`/etc/init.d/postgresql stop` and `pg_ctlcluster stop` both fail with **"pid file is invalid"** inside the FAB chroot because `/proc` is shared with the host. The PID file written by PostgreSQL contains the host-namespace PID, which the stop wrappers cannot validate against the chroot's `/proc`.

## Approach

Per @JedMeister's feedback on #347, this PR uses the existing fake `systemctl`/`service` infrastructure rather than calling `pg_ctl` directly from `conf/pgsql`. The call chain is:

```
conf/pgsql → systemctl start postgresql → fake service postgresql start → pg_ctl start
```

This keeps service management consistent with how other services (apache2, postfix, mariadb, etc.) are handled during builds.

## Changes

1. **`overlays/turnkey.d/systemd-chroot/usr/local/bin/service`** — new `postgresql)` case that resolves the PostgreSQL version at runtime and calls `pg_ctl` via `su postgres` for `start`, `stop`, and `restart`
2. **`conf/pgsql`** — replace all `/etc/init.d/postgresql` calls with `systemctl`; remove `--start` from `pg_createcluster` (start explicitly via `systemctl` instead); add `systemctl restart` after config changes so they take effect

## Test plan

- [ ] Full appliance build with PostgreSQL (e.g., NetBox) completes successfully
- [ ] PostgreSQL cluster creation, password config, user creation, and clean shutdown all work inside FAB chroot
- [ ] No regressions in other services managed by the fake `service` wrapper

Supersedes #347 (clean branch from `19.x-dev` with only the PostgreSQL fix).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conf/pgsql
+++ b/conf/pgsql
@@ -7,8 +7,8 @@ CONF_DIR=/etc/postgresql/${PGSQL_VER}/main
 
 # Recreate postgres and templates using UTF-8
 pg_dropcluster --stop ${PGSQL_VER} main || true
-pg_createcluster --start -e UTF-8 ${PGSQL_VER} main
-/etc/init.d/postgresql stop
+pg_createcluster -e UTF-8 ${PGSQL_VER} main
+systemctl start postgresql
 
 # Enable password encryption
 sed -i "/^#password_encryption = on/ s/#//" $CONF_DIR/postgresql.conf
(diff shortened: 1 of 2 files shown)


marcos-mendez:
alucinations

PR #351, opened 2026-04-21
Node nginx fixes

OnGle:


JedMeister:
Looks like the config was a bit messed up!

PR #355, opened 2026-05-13
IPv6 firewall webmin

JedMeister:
- <s>Add `libsocket6-perl` to base plan - required by Webmin for IPv6</s>
    - <s>ideally it should be added as a Webmin package depends, but here will do for now</s>
    - see: https://github.com/turnkeylinux/webmin/pull/14
    - (removed commit in this branch and force pushed)
- Write both IPv4 & IPv6 firewall rules.


marcos-mendez (approved):
NEAT!

PR #356, opened 2026-05-18
Webmin conf improvements

JedMeister:
- Write IPv6 firewall rules too.
- Enable HSTS by default
- Redirect HTTP to HTTPS (HTTP just failed before)
- Refactor logging setup - configure log locations in /var/log rather than making symlinks. - note webmin.log is still in default location so that needs a symlink still - however the file is now in /var/log/webmin and link is in /var/lib/webmin
- merge webmin log setup and the general webmin conf script into a single conf script (so can leverage the existing functionality in that)
- fail2ban tweaks (not specific to webmin and probably shouldn't be bundled here, but did the tweak at the same time, so here it is...)


OnGle (approved):
lgtm!

PR #357, opened 2026-05-22
Adminer (mostly...)

JedMeister:
This is built on top of #356 - so it may be best to review that first?

- adminer specific logging - common to all webservers
- adminer log rotation
- adminer fail2ban conf to protect adminer against brute force attacks - works with all webservers
- LigHTTPd specific adminer conf fixes (plus enable TKL configured ciphers by default on LigHTTPd)
- removed redundant fail2ban bugfixes from `conf/turnkey.d/fail2ban-fixes` (both fixed in Debian).
- Nginx related fixes/adjustments - refactor default config, enable php-fpm by default & fix adminer conf


OnGle (approved):
lgtm!

PR #358, opened 2026-05-29
Update TurnKey signing keys in 'master'

JedMeister:
Even though we use versioned branches now, I like to keep the keys up to date in master. That means that any instructions on where to download them can point to the master branch rather than needing to be updated for every release.

This PR adds the Trixie keys - apt and image signing - and the missing Bookworm image signing keys.

FYI the TurnKey keys are packaged for Trixie (`turnkey-keys`) so the repo signing keys here probably aren't as important moving forward, but probably best to include them...


OnGle (approved):
lgtm!

PR #359, opened 2026-06-01
Add postfix conf script port 25 check

JedMeister:
Adds a check to ensure that port 25 is free. Previously it would silently error and postfix had particularly unhelpful error message so it was a PITA to debug!

FWIW I had to `fab-investigate` my `root.patched`, install rsyslog, start it running, rerun postfix and read the resulting log. One of the downsides of systemd journal...


OnGle (approved):
lgtm!

marcos-mendez (approved):

PR #364, opened 2026-07-14
Webmin related updates

JedMeister:
Apologies for poor branch name and grab bag of changes...

- auto-expire webmin login\[1\]
- don't install `iptables-persistent` (causes issues if Webmin firewall start on boot enabled)
- install `webmin-systemd` package - manage systemd via webmin
- remove additional newlines in interfaces file
- ensure firewall config dir exists (not sure why but a recent build failed because it didn't?!)
- resolve boot time race condition when both ipv4 & ipv6 firewalls enabled via webmin

---

1. Previously login would never expire. Expiry can be disabled by checking the "remember me" checkbox on the login screen. And/or log in timeout can be adjusted or disabled via: 
   > Webmin >> Webmin Configuration >> Authentication >>  Auto-logout after X minutes of inactivity (under "Authentication options")


OnGle (approved):
lgtm!

PR #369, opened 2026-07-29
Move broken tkl-upgrade-php script out

### LINKED ISSUE #2023: Package tkl-upgrade-php script
It would be great to package the [`tkl-upgrade-php`](https://github.com/turnkeylinux/common/blob/551f9dc499bb8ff953531bd0c5e95b862e6dbf00/overlays/php/usr/local/bin/tkl-upgrade-php) script developed by @marcos-mendez (currently in the common PHP overlay). That would make it much easier for users to update to newer PHP versions as they become available. It could also warn the user if they are using a PHP version if it's no longer supported.

PHP provide details of PHP versions and their support status suitable for parsing here: https://www.php.net/releases/active

Note that while this issue is tagged with "LAMP" it applies to all appliances which include PHP.

JedMeister:
Seeing as the current iteration of `tkl-php-upgrade` is broken it's better to not include it at all. Besides it's better to package it rather than including it as an overlay script. That way it will be easy for TurnKey users to install & update.

For now, it has been moved to it's own repo here: https://github.com/turnkeylinux/tkl-upgrade-php

Once we get some time to circle back to it, it will be fixed and released as a package via our apt repo.

Even though the new tkl-php-upgrade repo does not build a package and is incomplete, now it's been moved to it's own repo the packaging work is now "in progress". So IMO:

Closes https://github.com/turnkeylinux/tracker/issues/2023

OTOH perhaps it should be left open and just tagged "in progress" - which I've just done regardless...

Also related to this show stopper issue (reason why it's being removed):
https://github.com/turnkeylinux/tracker/issues/2074


OnGle (approved):
Yeah agreed, this is the right move atm.

PR #372, opened 2026-08-06
19.0 common Fileserver updates.

JedMeister:
Cleaned things up a fair bit...

The only thing I wasn't sure about was the doc file? Open to suggestions...


OnGle (approved):
lgtm, docs seem good to me even if it's a little less than ideal.

PR #373, opened 2026-08-07
Add wsdd-daemon package

JedMeister:
The wsdd package (required for Windows to find fileserver via "network browsing") has been separated into 2 packages. They separated out the core functionality and the daemon. But in their infinite wisdom they kept the existing package name for the core/common functionality and (re)named the daemon package (required for previous functionality). Anyway, enough bitching from me...

Also FYI, wsdd didn't make it into trixie (an rc bug just before release). So I've rebuilt the Debian forky (testing) package source against trixie and pushed to our trixie main repo.


OnGle (approved):
lgtm!

PR #375, opened 2026-08-14
Bump TKL NodeJS landing page dep to resolve security issue

JedMeister:
Considering the context of our usage, this is unlikely to be an issue, but let's not use insecure deps anyway...

FYI the issue with body-parser < 1.20.6 is "denial of service when invalid limit value silently disables size enforcement".


OnGle (approved):
lgtm!

PR #377, opened 2026-08-18
Remove redundant LigHTTPd conf lines (modules already enabled elsewhere)

JedMeister:
`mod_redirect` & `mod_setenv` already enabled elsewhere, so they are redundant here - plus they were causing a deprecation warning. It was only a warning but may as well fix it now...


OnGle (approved):
lgtm!

PR #378, opened 2026-08-20
Modernize, bugfix and format bin/mysqlconf.py

JedMeister:
Sorry about all the changes...

I hit an edge case where this inithook errored and I'm pretty sure it was a race condition caused by this trying to start MySQL/MariaDB while the systemd service was also starting. Once I looked at it, there was so much legacy code that I thought the whole thing was worth updating - which in turn lead to more tidying up...

Changes of consequence:
- Replace `os.system()` with `subprocess.run()`.
- Use `os.makedirs()` & `shutil.chown()` to make & chown dirs (rather than bash commands via `os.system()`).
- <s>Add `--user=root` when starting MySQL/MariaDB (will fail to start as root unless explicit).</s>
- Use systemd to start/stop/check status - avoids race condition between inithook manual start with service. If this script is used at build time it will still work due to our "alternate" systemctl overlay script.
- Linting, typing and formatting.


OnGle (approved):
lgtm!

